PR #1, opened 2016-11-18
gsctl: First version

marians:
In case you want to test:

```
make
make install

gsctl help
gsctl info
gsctl login name@example.com
gsctl list organizations
gsctl list clusters
gsctl create kubeconfig [-c <cluster-id>]
gsctl list keypairs [-c <cluster-id>]
gsctl logout
```

### Related

~~https://github.com/giantswarm/gsclientgen/pull/1~~

### TODO

- [x] Add VERSION file for builder
- [x] Add a way to publish a release, including via homebrew
- [x] Change `gsclientgen` to branch `master` when gsclientgen/#1 is merged.


teemow:
My 2 cents. If somebody runs `make` it should build only the binary that is necessary for the user. If you'd like to create binaries for a release run `make bin-dist`. And `make install` is missing (we usually copy the binary to `/usr/local/bin` for convenience.

marians:
@teemow `make && make install` now work as wished.

commands/create.go
```diff
@@ -0,0 +1,190 @@
...
+	// TODO: Make ttl configurable
+	ttlHours := int32(3)
+	addKeyPairBody := gsclientgen.AddKeyPairBody{Description: cmdDescription, TtlHours: ttlHours}
+	keypairResponse, _, err := client.AddKeyPair(authHeader, cmdClusterID, addKeyPairBody)
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if keypairResponse.StatusCode == 10000 {
```
JosephSalisbury: This feels a bit magic - can we grab it from the generated client, or move it to a constant?

commands/login.go
```diff
@@ -0,0 +1,98 @@
...
+	loginResponse, apiResponse, err := client.UserLogin(email, requestBody)
+	if err != nil {
+		log.Fatal(err)
+	}
+	if loginResponse.StatusCode == 10000 {
+		// successful login
+		fmt.Println(color.GreenString("Successfully logged in"))
+		config.Config.Token = loginResponse.Data.Id
+		config.Config.Email = email
+	} else if loginResponse.StatusCode == 10010 {
```
JosephSalisbury: Same here with the status codes, and in a few other places

commands/ping.go
```diff
@@ -0,0 +1,43 @@
...
+func ping(cmd *cobra.Command, args []string) {
+	uri := "https://api.giantswarm.io/v1/ping"
+	start := time.Now()
+	resp, err := http.Get(uri)
+	if err != nil {
+		fmt.Println(color.RedString("API cannot be reached"))
+		fmt.Println(err.Error())
+		os.Exit(1)
+	} else {
+		defer resp.Body.Close()
```
JosephSalisbury: I'd elide this else, as you're already returning inside the `err != nil` block

so, like:
```
if err != nil {
   ...
   os.Exit(1)
}

defer resp.Body.Close()
...
```

config/config.go
```diff
@@ -0,0 +1,199 @@
...
+	yamlErr := yaml.Unmarshal(data, &myConfig)
+	if yamlErr != nil {
+		return nil, yamlErr
+	}
+
+	return myConfig, nil
+}
+
+// getKubeconfigPaths returns a slice of paths to known kubeconfig files.
+func getKubeconfigPaths(homeDir string) ([]string, error) {
```
JosephSalisbury: Your return signature is `([]string, error)`, but I can't actually see any errors being returned - can we simplify to just returning `[]string`?

util/kubectl.go
```diff
@@ -0,0 +1,118 @@
...
+	// name of the kubectl executable
+	binaryName string = "kubectl"
+
+	// url to intallatzion instructions
+	kubectlInstallURL string = "http://kubernetes.io/docs/user-guide/prereqs/"
+
+	kubectlWindowsInstallURL string = "https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md"
+)
+
+// CheckKubectl checks if kubectl exists, is silent if yes, reports error if not
```
JosephSalisbury: I'd make this a little more precise - it _prints_ an error if kubectl doesn't exist, and also exits

main.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// disable color on windows, as it is super slow
+	if runtime.GOOS == "windows" {
+		color.NoColor = true
+	}
+}
+
+func main() {
+	// top level commands
+	commands.RootCommand.AddCommand(
```
JosephSalisbury: With cobra based applications, a pattern I'm more familiar with is only the root command being exported, and the subcommands adding themselves in their own `init` - it took me a while to find this.

commands/create.go
```diff
@@ -0,0 +1,190 @@
...
+	if cmdDescription == "" {
+		return errors.New("No description given. Please use the -d/--description flag to set a description.")
+	}
+	return nil
+}
+
+func addKeypair(cmd *cobra.Command, args []string) {
+	client := gsclientgen.NewDefaultApi()
+	authHeader := "giantswarm " + config.Config.Token
+	// TODO: Make ttl configurable
```
JosephSalisbury: This is already done https://github.com/giantswarm/gsctl/pull/1/files#diff-d979e9548d0d641ce071ee98608da5faR132 here, so we might as well get it sorted

commands/create.go
marians: I don't see a way to provide this in the generated client. Any suggestion where to put a constant definition for these in gsclient?

main.go
xh3b4sd: + on internal sub command registration (registering internally with default values)
- on private/unexported commands (having godoc for the command is cool)

commands/ping.go
marians: check

commands/create.go
JosephSalisbury: If we can't stick it in the generated client, 'best' approach would probably be to have them defined in some other repository, and link it here.

Alternatively, make some constants in this project as required, and keep it simpler.

config/config.go
marians: check

commands/create.go
xh3b4sd: This comes probably from https://github.com/giantswarm/api-schema/blob/master/api_schema.go#L4. It feels a little bit odd that client specific status codes need to be handles in the command line tool. This separation makes no sense to me design wise. Anyway, instead of the magic number, you can use the apischema library.

commands/create.go
```diff
@@ -0,0 +1,189 @@
...
+		clientCertPath := util.StoreClientCertificate(config.ConfigDirPath, cmdClusterID, keypairResponse.Data.Id, keypairResponse.Data.ClientCertificateData)
+
+		clientKeyPath := util.StoreClientKey(config.ConfigDirPath, cmdClusterID, keypairResponse.Data.Id, keypairResponse.Data.ClientKeyData)
+
+		fmt.Println("Certificate and key files written to:")
+		fmt.Println(caCertPath)
+		fmt.Println(clientCertPath)
+		fmt.Println(clientKeyPath)
+
+		// TODO: Take this from the cluster object
```
xh3b4sd: Please create an issue for this if it will not be addressed in this PR.

commands/create.go
```diff
@@ -0,0 +1,189 @@
...
+
+		// final success message
+		color.Green("kubectl is set up. Check it using this command:\n\n")
+		color.Yellow("    kubectl cluster-info\n\n")
+		color.Green("Whenever you want to switch to using this context:\n\n")
+		color.Yellow("    kubectl config set-context giantswarm-%s\n\n", cmdClusterID)
+
+	} else {
+		fmt.Printf("Unhandled response code: %v", keypairResponse.StatusCode)
+		fmt.Printf("Status text: %v", keypairResponse.StatusText)
```
xh3b4sd: I find it weird to use different libraries to do one thing. You print information using `fmt` and `color`. There should only be one way to do it.

commands/info.go
```diff
@@ -0,0 +1,76 @@
...
+		Use:   "info",
+		Short: "Print some information",
+		Long:  `Prints information that might help you get out of trouble`,
+		Run:   printInfo,
+	}
+)
+
+// printInfo prints some information on the current user and configuration
+func printInfo(cmd *cobra.Command, args []string) {
+
```
xh3b4sd: This empty line can be removed.

main.go
marians: check

commands/info.go
```diff
@@ -0,0 +1,76 @@
...
+			output = append(output, color.YellowString("Auth token:")+"|"+color.CyanString(config.Config.Token))
+		} else {
+			output = append(output, color.YellowString("Auth token:")+"|n/a")
+		}
+	}
+
+	output = append(output, color.YellowString("%s version:", config.ProgramName)+"|"+config.Version)
+	output = append(output, color.YellowString("%s build:", config.ProgramName)+"|"+config.BuildDate)
+
+	output = append(output, color.YellowString("Config path:")+"|"+color.CyanString(config.ConfigFilePath))
```
xh3b4sd: There is probably a reason but it looks like we are doing the table generation wrong. You may want to create an issue to improve this, if not already done.

config/config.go
```diff
@@ -0,0 +1,199 @@
...
+	SystemUser *user.User
+)
+
+func init() {
+
+	SystemUser, userErr := user.Current()
+	checkErr(userErr)
+
+	ConfigDirPath = SystemUser.HomeDir + string(os.PathSeparator) + "." + ProgramName
+	ConfigFilePath = ConfigDirPath + string(os.PathSeparator) + ConfigFileName
```
xh3b4sd: You can use https://golang.org/pkg/path/filepath/ to join file paths.

config/config.go
```diff
@@ -0,0 +1,199 @@
...
+
+func checkErr(err error) {
+	if err != nil {
+		panic(err.Error())
+	}
+}
+
+// WriteToFile writes the configuration data to a YAML file
+func WriteToFile() error {
+
```
xh3b4sd: Empty line.

config/config.go
```diff
@@ -0,0 +1,199 @@
...
+	if yamlErr != nil {
+		return nil, yamlErr
+	}
+
+	return myConfig, nil
+}
+
+// getKubeconfigPaths returns a slice of paths to known kubeconfig files
+func getKubeconfigPaths(homeDir string) []string {
+
```
xh3b4sd: Empty line.

config/config.go
```diff
@@ -0,0 +1,199 @@
...
+	kubeconfigEnv := os.Getenv("KUBECONFIG")
+	if kubeconfigEnv != "" {
+		// KUBECONFIG is set.
+		// Now check all the paths included for file existence
+
+		// "windows" is really the value for the windows platform.
+		// No constant for that in runtime.
+		pathSep := ":"
+		if runtime.GOOS == "windows" {
+			pathSep = ";"
```
xh3b4sd: You can use https://golang.org/pkg/path/filepath/#pkg-constants.

marians:
@JosephSalisbury Thanks for your review! Apart from the [response code constants part](https://github.com/giantswarm/gsctl/pull/1#discussion_r89518072), I have acted on all your suggestions.

config/config.go
```diff
@@ -0,0 +1,199 @@
...
+					clusterIDs = append(clusterIDs, cluster.Id)
+				}
+			}
+			if len(clusterIDs) == 1 {
+				return clusterIDs[0], nil
+			}
+			return "", nil
+		}
+	}
+	return "", errors.New(orgsResponse.StatusText)
```
xh3b4sd: A general note on error handling. We use https://github.com/juju/errgo in most places which is useful when something goes wrong. In your current code you do not know where the code flows along when something breaks. Then debugging is a huge pain, especially for people who are not familiar with the code base. The approach that proved quite good was to have a `error.go` in each package that has to deal with errors. See e.g. https://github.com/giantswarm/kubernetesd/blob/master/service/creator/error.go. Errors being defined there can then be used like in the following places. 

- https://github.com/giantswarm/kubernetesd/blob/master/service/creator/creator.go#L76
- https://github.com/giantswarm/kubernetesd/blob/master/service/creator/creator.go#L106

You may want to create an issue for this.

util/certs.go
```diff
@@ -0,0 +1,57 @@
+package util
```
xh3b4sd: I am not a fan of this `util` package. There should be separated packages doing one thing well.

JosephSalisbury:
rock and roll!

util/kubectl.go
```diff
@@ -0,0 +1,118 @@
...
+	cmd := exec.Command(binaryName,
+		"config", "set-cluster",
+		clusterName,
+		serverArgument,
+		certificateAuthorityArgument)
+	if err := cmd.Run(); err != nil {
+		fmt.Println(color.RedString("Could not set cluster using 'kubectl config set-cluster ...'"))
+		fmt.Println("Error:")
+		fmt.Println(err)
+		os.Exit(1)
```
xh3b4sd: I think it is wrong to terminate the program from within the deeps of the code base. A program should be initialized, executed and shutdown in the same place.

marians:
@xh3b4sd I am through with your feedback. Thanks for the extensive review! I would like to leave refactoring of the `util` package for a later step.

Also @JosephSalisbury If you are fine with the changes, please hit the Approve button. Thanks!

Issue #4, opened 2016-11-25
gsctl version


marians:
The CLI is lacking a command to quickly check the version.

Seems like this is expected under `gsctl --version` (in addition to `gsctl version`).

marians:
Closed with 0.2.0

PR #6, opened 2016-11-25
Add activity tracking, List output optimization and color opt-out, add version command

marians:
This PR adds two unrelated changes:

## Activity Tracking

Add tracking of user actions ("activities") via request headers, which helps us to understand what users do with the CLI. This submits two request headers to the API:

(1) a request ID, which will also be helpful when debugging requests in the future.

(2) the complete command line. Part (2) can be deactivated using the `GSCTL_DISABLE_CMDLINE_TRACKING` environment variable (see README).

## Output improvements

These changes are based on user feedback (thanks @teemow!)

- Colors can now be turned off by setting the environment variable `GSCTL_DISABLE_COLORS`.
- Table output for the commands `gsctl list clusters`, `gsctl list organizations`, and `gsctl list keypairs` has been changed so that only headers are printed in color, while data rows use the default terminal color.
- To make table column headers easier to distinguish from data rows, even with colors disabled, they are now CAPITALIZED.

## Version command

- The `gsctl version` command has been added, with the purpose to simply output the version number (and build date when used with `--verbose`). See #4.

## TODO

- [ ] When https://github.com/giantswarm/gsclientgen/pull/4 is merged, set the branch used back to master

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,15 +28,15 @@ get-deps: .gobuild
 .gobuild:
 	@mkdir -p $(GS_PATH)
 	@rm -f $(GS_PATH)/$(PROJECT) && cd "$(GS_PATH)" && ln -s ../../../.. $(PROJECT)
-	#builder get dep -b branch-name https://github.com/giantswarm/gsclientgen.git $(GS_PATH)/gsclientgen
-	go get -v github.com/giantswarm/gsclientgen
-	go get -v github.com/giantswarm/api-schema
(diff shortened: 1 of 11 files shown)


oponder:
LGTM! Seems to WFM also!

Maybe for a next topic, look a bit at error handling / error messages to provide something friendlier than this to the users:

```
$ ./gsctl list keypairs -c asdf
2016/11/28 16:13:58 unexpected end of JSON input
```

marians:
@oponder Noted this example in a [new issue](https://github.com/giantswarm/gsctl/issues/8). I am aware that this sort of improvement is needed in more than just this place.

PR #11, opened 2016-12-05
added api endpoint flag

xh3b4sd:
This PR makes the GS API endpoint configurable. 

RFR @JosephSalisbury @marians

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -10,6 +10,7 @@ import (
 var RootCommand = &cobra.Command{Use: config.ProgramName}
 
 func init() {
+	RootCommand.PersistentFlags().StringVarP(&cmdAPIEndpoint, "api-endpoint", "", config.APIEndpoint, "The swagger base path, to access the API")
 	RootCommand.PersistentFlags().StringVarP(&cmdToken, "auth-token", "", "", "Authorization token to use for one command execution")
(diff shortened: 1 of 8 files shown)


commands/root.go
```diff
@@ -10,6 +10,7 @@ import (
 var RootCommand = &cobra.Command{Use: config.ProgramName}
 
 func init() {
+	RootCommand.PersistentFlags().StringVarP(&cmdAPIEndpoint, "api-endpoint", "", config.APIEndpoint, "The swagger base path, to access the API")
```
JosephSalisbury: I don't see why swagger is important here - this is the URL base path, right?

config/config.go
```diff
@@ -31,6 +31,9 @@ const (
 )
 
 var (
+	// APIEndpoint is the swagger base path, to access the API.
```
JosephSalisbury: Same here

marians:
I agree with what @JosephSalisbury commented.

One additional question: where is the PR introducing `gsclientgen.NewDefaultApiWithBasePath()`?

commands/root.go
xh3b4sd: Yes.

xh3b4sd:
@marian `gsclientgen.NewDefaultApiWithBasePath` is generated by swagger and already there.

Issue #9, opened 2016-12-01
Makefile issues


teemow:
While trying to build a new version `0.2` after I've build `0.1` I ran into the following issues.

* Updated git repo
* Ran `make install` resulted in only copying the old binary
```
➜ make install
cp .gobuild/bin/gsctl-linux-amd64 /usr/local/bin/gsctl
```
* Running `make` only fetched the dependencies again:
```
make
#builder get dep -b branch-name https://github.com/giantswarm/gsclientgen.git "/home/teemow/projects/giantswarm/gsctl/.gobuild/src/github.com/giantswarm"/gsclientgen
go get -v github.com/giantswarm/gsclientgen
go get -v github.com/bradfitz/slice
go get -v github.com/fatih/color
go get -v github.com/giantswarm/api-schema
go get -v github.com/giantswarm/columnize
github.com/giantswarm/columnize (download)
github.com/giantswarm/columnize
go get -v github.com/go-resty/resty
go get -v github.com/howeyc/gopass
go get -v github.com/inconshreveable/mousetrap
go get -v github.com/spf13/cobra/cobra
go get -v gopkg.in/yaml.v2
```
* Running `make build` said there is nothing to do (because the binary exists but you should actually check if the source files have changed)
```
➜ make build
make: Für das Ziel „build“ ist nichts zu tun.
```
* Running `make clean`
* Running `make install` again didn't fetch the dependencies again:
```
➜ make install
docker run --rm -v /home/teemow/projects/giantswarm/gsctl:/usr/code -w /usr/code \
        -e GOPATH=/usr/code/.gobuild -e GOOS=linux -e GOARCH=amd64 -e CGO_ENABLED=0 \
        golang:1.7.3 go build -a -installsuffix cgo -o .gobuild/bin/gsctl-linux-amd64 \
        -ldflags "-X 'github.com/giantswarm/gsctl/config.Version=0.2.0+git' -X 'github.com/giantswarm/gsctl/config.BuildDate=2016-12-01T10:55:49Z' -X 'github.com/giantswarm/gsctl/config.Commit=071c0d1463'"
main.go:7:2: cannot find package "github.com/fatih/color" in any of:
        /usr/local/go/src/github.com/fatih/color (from $GOROOT)
        /usr/code/.gobuild/src/github.com/fatih/color (from $GOPATH)
main.go:8:2: cannot find package "github.com/giantswarm/columnize" in any of:
[...]

PR #12, opened 2016-12-05
Ensure re-build

marians:
- `build` should compile again when sources changed
- `install` should build again when sources changed, and ensure
dependencies


marians:
@teemow This should solve the issues you mentioned in https://github.com/giantswarm/gsctl/issues/9

marians:
@oponder Seems like @teemow is too busy for this review. Can you jump in?

PR #13, opened 2016-12-15
Simpler Build, vendored dependencies

marians:
Don't let the massive list of additions fool you. ;-)

This PR is about using standard tools for vendored dependencies (`glide` in this case) and removing the need to tweak $GOPATH in order to create a build and no need to have `builder` in order to fetch dependencies.


.travis.yml
```diff
@@ -1,5 +1,3 @@
-sudo: required # needed for docker
```
marians: It turns out that Travis doesn't need this to run a docker container.

As a result, the build process is much faster, from > 3 minutes to slightly over one minute.

marians:
After these dependency changes, but no code changes, and also going to GO 1.7.4, I have this problem:

image

image

The problem: The column headers aren't in the right position.

Release 0.2.0:

image

marians:
Output glitches are fixed through https://github.com/giantswarm/columnize/pull/2

glide.lock
```diff
@@ -8,7 +8,8 @@ imports:
 - name: github.com/giantswarm/api-schema
   version: f8468345e621ea9f5af6965919449dbbe827a841
 - name: github.com/giantswarm/columnize
-  version: 822fe83f8e3222d9cee96ad5d10aad8e00dccc8a
+  version: 677ded9e47e75ca49bd2541794a7281a49a2f17b
+  repo: https://github.com/giantswarm/columnize.git
```
JosephSalisbury: I don't like this - we shouldn't need to specify repository URLs

glide.yaml
```diff
@@ -6,7 +6,8 @@ import:
 - package: github.com/giantswarm/api-schema
   version: ^0.7.1
 - package: github.com/giantswarm/columnize
-  version: fix-length-calculation
+  version: =2.0.2
+  repo: https://github.com/giantswarm/columnize.git
```
JosephSalisbury: And here too

glide.lock
marians: Well, YOU told me to

marians: glide.lock is generated and only refects what's specified in glide.yaml...

marians:
@JosephSalisbury Can you maybe tell me how I can make this PR be based on `master`? The branch `make-sources` is already merged into master.

JosephSalisbury:
@marians `Edit` button in the top right, and then change the base branch

glide.yaml
marians: Removed

marians:
@JosephSalisbury The requested change has been made

marians:
Thanks man!

JosephSalisbury:
No worries <3

PR #18, opened 2017-01-05
Use mac keychain

marians:
This change only affects Mac users.

On `gsctl login`, we now look for a password in the user's Keychain. If it's there, we use this instead of prompting for a password. If there is no matching entry in the Keychain, we prompt for password input.

After a successful login, we store the password in the Keychain for future logins (see above).

With -v we also print "Using password from Keychain" when this is the case.

Questions:

- What happens if the keychain is locked? Is the user asked for the Keychain master password (as expected)?
  - yes, in a GUI prompt dialog
- Should we make it more visible that we are reading and writing the password from/to the Keychain (not only when using -v)?
- Should we allow users to opt out of storing the password in the Keychain?
- What happens if the user changes the password e. g. in happa? How will the stored password be updated?


puja108:
> What happens if the keychain is locked? Is the user asked for the Keychain master password (as expected)?

AFAIK yes

> Should we make it more visible that we are reading and writing the password from/to the Keychain (not only when using -v)?
> Should we allow users to opt out of storing the password in the Keychain?

IMHO these questions are connected. I'd say we ask for confirmation (opt-in) and then the user knows so we don't need to tell him each time. So yes, it should not be forced.

> What happens if the user changes the password e. g. in happa? How will the stored password be updated?

I think that is also handled when keychain sees it does not work, like when you change the WiFi password or sth. It can then be saved again overwriting the old one I guess?

marians:
> What happens if the keychain is locked? Is the user asked for the Keychain master password (as expected)?

Yes, I checked it now.

marians:
Currently this only builds on Mac OS, so isn't working in our Linux cross-compile workflow. We might look into the suggested made at https://github.com/lunixbochs/go-keychain/issues/4 at some point. No urgency.

PR #27, opened 2017-03-07
Use Go 1.8 for builds

marians:
This PR switches our build (and Travis tests) to Golang 1.8. For builds, the smaller `golang:1.8-alpine` image is now used instead of the standard one.


marians:
Thanks boys!

PR #30, opened 2017-03-14
Printing raw api response in case of unknown error

marians:
This should help with API debugging

Ping @webwurst


marians:
Closing this, as it didn't help and the solution wasn't so great.

Issue #2, opened 2016-11-24
Unify internal error handling


marians:
@xh3b4sd suggests in https://github.com/giantswarm/gsctl/pull/1/files#r89526585:

A general note on error handling. We use https://github.com/juju/errgo in most places which is useful when something goes wrong. In your current code you do not know where the code flows along when something breaks. Then debugging is a huge pain, especially for people who are not familiar with the code base. The approach that proved quite good was to have a `error.go` in each package that has to deal with errors. See e.g. https://github.com/giantswarm/kubernetesd/blob/master/service/creator/error.go. Errors being defined there can then be used like in the following places. 

- https://github.com/giantswarm/kubernetesd/blob/master/service/creator/creator.go#L76
- https://github.com/giantswarm/kubernetesd/blob/master/service/creator/creator.go#L106

marians:
With #32, error handling has been unified throughout the code base. There is room for improvement, but for now this is okay.

PR #32, opened 2017-03-17
Unified handling of unknown API responses

marians:
This change introduces a `dumpAPIResponse()` function that is used to print verbose information in case of unecpected API responses.

Example (with en empty response body):

```
API request/response details:
Operation:      GetKeyPairs (GET https://api.giantswarm.io/v3/clusters/l8d/key-pairs/)
Status:         403 Forbidden
Response body:
```


marians:
related to #2

PR #36, opened 2017-04-04
Adding some tests

marians:
@rossf7 @JosephSalisbury This is more or less my first venture into writing tests in Go. It's a start and the tests only cover minor functions. Call it low hanging fruit. I'd be happy for a review/feedback from one of the two of you. Thanks!


Makefile
```diff
@@ -67,6 +67,10 @@ build/bin/$(BIN)-windows-amd64:
 		golang:$(GOVERSION) go build -a -installsuffix cgo -o build/bin/$(BIN)-windows-amd64 \
 		-ldflags "-X 'github.com/giantswarm/gsctl/config.Version=$(VERSION)' -X 'github.com/giantswarm/gsctl/config.BuildDate=$(BUILDDATE)' -X 'github.com/giantswarm/gsctl/config.Commit=$(COMMIT)'"
 
+# run unittests
+gotest:
+	go test -cover ./util ./config
```
rossf7: I'd used $(glide novendor) here rather than specifying the packages. There is also a change to Golang in progress to not run vendor tests which is a better default.

https://go-review.googlesource.com/c/38745/

JosephSalisbury: The changes to pattern matching aren't out yet, i'd use `$(glide novendor)`

util/kubectl_test.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+import "testing"
+
+func TestCheckKubectl(t *testing.T) {
+	result := CheckKubectl()
```
rossf7: This test seems a bit strange to me as looks like its testing whether kubectl is installed. There could also be problems running it on Circle CI.

marians: How about making `kubectl` available in our CircleCI build context? Would that be too much?

Makefile
marians: ✅

util/kubectl_test.go
rossf7: Yes you can install packages in the circle.yml so that should work. I think that makes sense especially if you want to write more tests that need kubectl.

https://circleci.com/docs/1.0/installing-custom-software/

rossf7: Ah yes I just noticed that you're using Travis not Circle! That will definitely work then as I've installed packages on Travis before :)

marians:
I managed to install glide in the Travis build environment and added `make gotest` to the tests executed. 

For some reason, `go test -cover $(glide novendor)` just results in `go test -cover`, which isn't what I'd expect.

See https://travis-ci.org/giantswarm/gsctl/builds/218862193#L480-L482

Any idea, releng folks?

marians:
Oh, I forgot `shell` in the Makefile...

marians:
@rossf7 Should have stuff sorted out. Okay now?

PR #42, opened 2017-04-18
Adding tests

marians:
This PR brings to the table:

- breaking down some command files into smaller chunks for better overview
- adding basic tests for `create keypair`, `create kubeconfig`, `list clusters`, `list_keypairs`

So far the tests ensure execution of the most commonly used paths in the above commands. API responses are mocked using `httptest.Server`.

Future work might include
- Breaking out the functionality from functions like `createKubeconfig()` into one/several that return `string, error`.
- Ensuring that no locally existing config file is used
- Testing more different cases

Fun fact: Atom's Go auto import feature is prone to import the wrong library. I was already using `github.com/giantswarm/columnize` in the project. However after refactoring, `columnize` has been imported as the commonly known `github.com/ryanuber/columnize`.


marians:
The latest codecov report is here: https://codecov.io/gh/giantswarm/gsctl/pull/42

coverage.txt
```diff
@@ -1,68 +1,361 @@
...
-github.com/giantswarm/gsctl/config/config.go:178.38,179.33 1 0
-github.com/giantswarm/gsctl/config/config.go:179.33,181.46 2 0
-github.com/giantswarm/gsctl/config/config.go:190.4,190.28 1 0
-github.com/giantswarm/gsctl/config/config.go:193.4,193.18 1 0
-github.com/giantswarm/gsctl/config/config.go:181.46,183.19 2 0
-github.com/giantswarm/gsctl/config/config.go:186.5,186.60 1 0
-github.com/giantswarm/gsctl/config/config.go:183.19,185.6 1 0
-github.com/giantswarm/gsctl/config/config.go:186.60,188.6 1 0
-github.com/giantswarm/gsctl/config/config.go:190.28,192.5 1 0
 mode: atomic
```
marians: This is necessary for code coverage reports. Codecov.io generates a before/after comparison for a PR based on the change of this. It's generated using `./test_coverage.sh`.

PR #49, opened 2017-05-03
Testdata

marians:
A bit of house keeping.

Test data is supposed to be kept in `testdata` folders in Go.

While at it I removed and edited bits, to make the test files work in current reality.


marians:
Reminder-RFR for @oponder

PR #50, opened 2017-05-05
Configurable client for faster timeouts

marians:
This PR adds a way to configure the API client for

- quicker timeouts where appropriate, e. g. when listing organizations
- Setting a user agent string

This is applied for all API requests.

The motivation were lots of timeouts when using our current leaseweb-g8s backend, especially in commands like `list clusters`, where multiple requests are made.

Technically this could have been part of `gsclientgen`, but in fact I don't want to add code there, as all code in that repository is generated by swagger.


commands/list_clusters.go
```diff
@@ -43,9 +44,14 @@ func checkListClusters(cmd *cobra.Command, args []string) error {
 
 // list all clusters the user has access to
 func listClusters(cmd *cobra.Command, args []string) {
-	client := gsclientgen.NewDefaultApiWithBasePath(cmdAPIEndpoint)
+	clientConfig := client.Configuration{
+		Endpoint:  cmdAPIEndpoint,
+		Timeout:   3 * time.Second,
```
xh3b4sd: What is actually the reasoning behind the very different timeouts?

client/client_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestTimeout(t *testing.T) {
+	clientConfig := Configuration{
+		Endpoint: "http://httpbin.org/delay/2?",
```
marians: I think depending on the network for tests should be fine. But I will look at the alternative. Having it totally internal is definitely better.

commands/list_clusters.go
marians: I want to fail fast in situations where multiple GET calls are made in sequence. E. g. when listing clusters, we first fetch all the organizations (memberships). Then we look for clusters in each org. This failing after one minute (as it happened just recently with flapping API behaviour) is just a really bad waste of time for the end user. I'd rather fail after a few seconds and let the user retry.

In cases of transactions (POST, PUT, PATCH) I'd like to tolerate longer timeouts, as we expect a little bit more work to happen on the other end, and retrying might not be as easy (unless transactions are implemented for that route).

The next step of course will be doing the retrying automatically, so the user doesn't have to bother at all in most cases.

It's a tiny step, but I see is as a little piece in a bigger resiliance picture.

client/client_test.go
```diff
@@ -0,0 +1,35 @@
...
+		Timeout:  1 * time.Second,
+	}
+	apiClient := NewClient(clientConfig)
+	_, _, err := apiClient.GetUserOrganizations("foo", "foo", "foo", "foo")
+	if err == nil || !strings.Contains(err.Error(), "Client.Timeout exceeded") {
```
marians: I'll give that a try

marians: Unfortunately, the error type isn't specific enough: `*url.Error`.

commands/create_cluster.go
```diff
@@ -335,8 +337,13 @@ func addCluster(cmd *cobra.Command, args []string) {
 			// command line flag overwrites
 			authHeader = "giantswarm " + cmdToken
 		}
-		client := gsclientgen.NewDefaultApiWithBasePath(cmdAPIEndpoint)
-		responseBody, apiResponse, _ := client.AddCluster(authHeader, addClusterBody, requestIDHeader, createClusterActivityName, cmdLine)
+		clientConfig := client.Configuration{
+			Endpoint:  cmdAPIEndpoint,
+			Timeout:   60 * time.Second,
```
marians: Done

marians:
@xh3b4sd Thanks for the review! Happy for a re-check.

client/client_test.go
xh3b4sd: When it is a URL error, we can check if it is an timeout error. See https://golang.org/pkg/net/url/#Error.Timeout.

client/client_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"time"
+)
+
+func TestTimeout(t *testing.T) {
+	// Our test server.
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// enforce a timeout longer than the client's
+		time.Sleep(3 * time.Second)
+		fmt.Fprintln(w, "Hello")
+	}))
```
xh3b4sd: I like that way more than before. Cool.

marians:
Changed error type checking.

PR #52, opened 2017-05-09
Removing selected organization and cluster

marians:
In `gsctl info` there were these two empty lines:

```
Selected organization:
Selected cluster:
```

and some places in the code actually checked for these settings, but apart from manually editing the config file it wasn't possible to set these. And I'm currently not sure how setting these should happen, so I'd rather remove this source of confusion for end users.


coverage.txt
```diff
@@ -265,28 +257,26 @@ github.com/giantswarm/gsctl/commands/create_keypair.go:43.2,43.24 1 1
...
+github.com/giantswarm/gsctl/commands/create_cluster.go:306.17,308.4 1 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:303.2,303.16 1 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:303.16,306.17 3 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:295.81,299.4 3 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:293.28,295.81 1 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:293.2,293.28 1 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:287.3,287.13 1 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:283.4,286.4 2 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:280.29,283.4 2 0
+github.com/giantswarm/gsctl/commands/create_cluster.go:278.28,280.29 2 0
```
xh3b4sd: IMO this file should not be committed.

marians: It has to be committed, otherwise there can't be test coverage comparison before/after by codecov.

xh3b4sd: I couldn't find either evidence. Do you have a source? The example go project in the codecov documentation overwrites the `coverage.txt` on each test run. This file is only used to upload the report. The comparison is done by codecove internally. IMHO this file should not be committed because it is also super irrelevant AFAIK. Maybe something changed meanwhile, but then provide please some docs about this because I couldn't find it.

marians: I'll give it a test. Maybe I got things wrong. More than just a possibility.

Issue #51, opened 2017-05-05
Type error with size_gb for cluster nodes


rossf7:
I hit this error getting kubeconfig for a test cluster. I'm using an awsgsctl alias to connect to the AWS endpoint.

```
awsgsctl create kubeconfig --cluster tvanc --auth-token ed477dfa-3667-43e9-bbed-3989cfd18ca5
Could not fetch details for cluster ID 'tvanc'
Error: json: cannot unmarshal number 7.5 into Go struct field V4NodeDefinitionMemory.size_gb of type int32
API request/response details:
Operation:      GetCluster (GET https://api-aws.giantswarm.io/v4/clusters/tvanc/)
Status:         200 OK
Response body:  {"api_endpoint":"https://api.tvanc.g8s.eu-west-1.adidas.aws.giantswarm.io","create_date":"2017-05-05T13:13:53.303618742Z","id":"tvanc","name":"ross","owner":"giantswarm","kubernetes_version":"v1.5.2_coreos.2","aws":{},"masters":[{"cpu":{"cores":2},"memory":{"size_gb":7.5},"storage":{"size_gb":32}}],"workers":[{"cpu":{"cores":2},"memory":{"size_gb":7.5},"storage":{"size_gb":32},"aws":{"instance_type":"m3.large"}},{"cpu":{"cores":2},"memory":{"size_gb":7.5},"storage":{"size_gb":32},"aws":{"instance_type":"m3.large"}}]}
giantswarm/kubernetesd - [update-vendor] » gsctl --version
```

xh3b4sd:
The version output is stripped. I was checking, and this should actually work. See https://github.com/giantswarm/gsclientgen/pull/10. Do you maybe need to rebuild the latest version?

rossf7:
I'm running the latest release but not master. I'll re-test with master.

rossf7:
I've retested with master and I'm still getting type errors with SizeGB.

```
awsgsctl create kubeconfig --cluster=9619m                                                                                                                 (g8s-aws/giantswarm)
Could not fetch details for cluster ID '9619m'
Error: json: cannot unmarshal number 7.5 into Go struct field V4NodeDefinitionMemory.size_gb of type int32
API request/response details:
Operation:      GetCluster (GET https://api-aws.giantswarm.io/v4/clusters/9619m/)
Status:         200 OK
[...]

PR #53, opened 2017-05-09
Changing SizeGB to float32

marians:
Related:

https://github.com/giantswarm/gsctl/issues/51
https://github.com/giantswarm/gsclientgen/pull/11

### TODO:

- [x] wait for merging of https://github.com/giantswarm/gsclientgen/pull/11
- [ ] change back gsclientgen version to master


marians:
@rossf7 Could you give this branch a try? `make clean && make && sudo make install` should do the job.

rossf7:
Thanks Marian, that works great  <3

```
awsgsctl create kubeconfig --cluster=lkv2p                                                                                                                    (g8s-aws/giantswarm)
Creating new key-pair…
New key-pair created with ID 4d821dda8… and expiry of 720 hours
Certificate and key files written to:
/Users/ross/.gsctl/certs/lkv2p-ca.crt
/Users/ross/.gsctl/certs/lkv2p-4d821dda83-client.crt
/Users/ross/.gsctl/certs/lkv2p-4d821dda83-client.key
Switched to kubectl context 'giantswarm-lkv2p'

kubectl is set up. Check it using this command:

    kubectl cluster-info

Whenever you want to switch to using this context:

    kubectl config set-context giantswarm-lkv2p
```

commands/create_cluster.go
```diff
@@ -23,11 +23,11 @@ type cpuDefinition struct {
 }
 
 type memoryDefinition struct {
-	SizeGB int `yaml:"size_gb,omitempty"`
+	SizeGB float32 `yaml:"size_gb,omitempty"`
 }
 
 type storageDefinition struct {
-	SizeGB int `yaml:"size_gb,omitempty"`
+	SizeGB float32 `yaml:"size_gb,omitempty"`
```
xh3b4sd: This is a personal thing. Would like to use `float64` for simplicity. We do not need to count bytes in the client. The API also uses `float64`. I feel like `float64` is just more common when you do not care about memory allocation.

marians: Hmm. Not convinced. I expect a maximum of two or three digits after the comma in practice. Why not use the leaner type then?

xh3b4sd: IMO the code is more easy when you only have to deal with one kind of types. At some point you have to type assert and cast information back and forth. Then it gets super messy because you have to use all kinds of different methods for 8, 16, 32 and 64 bits. It makes the code only more complicated IMO. Enough said. Do what you think is right.

marians:
Regarding the float32 type: I forgot to mention here that the choice has been made in generated code in https://github.com/giantswarm/gsclientgen/pull/11. swagger-codegen translates the JSON-schema type "number" to float32.

PR #54, opened 2017-05-10
Ping refactoring, testing, sending user-agent header

marians:
This PR

- ensures that ping requests send the user-agent header
- makes the ping function better testable by separating logic from output handling
- adds a unit test

### DIFF
--- a/commands/ping.go
+++ b/commands/ping.go
@@ -32,7 +32,7 @@ func init() {
 func runPingCommand(cmd *cobra.Command, args []string) {
 	duration, err := ping(cmdAPIEndpoint)
 	if err != nil {
-		fmt.Println(color.RedString("API connection is bad"))
+		fmt.Println(color.RedString("Could not reach API"))
 		fmt.Println(err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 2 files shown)


commands/ping.go
```diff
@@ -17,41 +19,65 @@ var (
...
-// ping checks the API connections
-func ping(cmd *cobra.Command, args []string) {
-	u, err := url.Parse(cmdAPIEndpoint)
+// runPingCommand executes the ping() function
+// and prints output in a user-friendly way
+func runPingCommand(cmd *cobra.Command, args []string) {
+	duration, err := ping(cmdAPIEndpoint)
 	if err != nil {
-		fmt.Println(err.Error())
+		fmt.Println(color.RedString("API connection is bad"))
```
oponder: I don't really like "API connection is bad" from a UX point of view.

Can we offer some more information here? Or at least something more concrete like "Could not reach API" ? It's so very subjective to hear something is bad.. I am left wondering what it means.

marians: Currently the user gets the errors details as an additional line:

```
API connection is bad
Get http://nonexisting/v1/ping: dial tcp: lookup nonexisting on 8.8.8.8:53: no such host
```

I'll happily change the wording to "Could not reach API".

Issue #39, opened 2017-04-12
Move config to ~/.config/gsctl


marians:
It seems `~/.config` has become a default place for configuration sub-directories. That's nice, because it avoids too much clutter in the home directory.

We should move our config there, to, in a sub-dir `~/.config/gsctl`.

marians:
Done

PR #55, opened 2017-05-10
Config refactoring

marians:
In this PR I'm introducing viper as a means to deal with settings to be set via defaults, config file, environment variables, and flags.

Changes affecting users:

- The config path changes from `$HOME/.gsctl` to `$HOME/.config/gsctl` (https://github.com/giantswarm/gsctl/issues/39), to avoid clutter of a user's home dir. Existing config is migrated automatically.

Additionally:
- unit tests for the `config` package are added.
- tests no longer drop certificates in the user's actual certs dir (towards #56)

More fruits of this will hopefully be harvested in future PRs.


config/config.go
```diff
@@ -174,19 +222,26 @@ func getKubeconfigPaths(homeDir string) []string {
...
+	}
+	apiClient := client.NewClient(clientConfig)
+
 	authHeader := "giantswarm " + Config.Token
-	orgsResponse, _, err := client.GetUserOrganizations(authHeader, requestIDHeader, activityName, cmdLine)
+	orgsResponse, _, err := apiClient.GetUserOrganizations(authHeader, requestIDHeader, activityName, cmdLine)
 	if err != nil {
 		return "", err
 	}
 	if orgsResponse.StatusCode == 10000 {
```
rossf7: Why is this status code 10000? I'm confused about what's happening here.

config/config.go
```diff
@@ -202,3 +257,26 @@ func GetDefaultCluster(requestIDHeader, activityName, cmdLine, cmdAPIEndpoint st
...
+	}
+
+	_, err = os.Stat(LegacyConfigDirPath)
+	if os.IsNotExist(err) {
+		// old config dir does not exist
+		return nil
+	}
+
+	// ensure ~/.config exists
+	os.MkdirAll(path.Dir(DefaultConfigDirPath), 0700)
```
rossf7: Bit of a nitpick but os.MkdirAll can return an error. Should be OK as its the home dir but might be good to check for it.

config/config_test.go
```diff
@@ -1,8 +1,135 @@
...
+	dir := tempDir()
+	defer os.RemoveAll(dir)
+	filePath := path.Join(dir, ConfigFileName+"."+ConfigFileType)
+
+	// write dummy config
+	file, fileErr := os.Create(filePath)
+	if fileErr != nil {
+		t.Error(fileErr)
+	}
+	file.WriteString("email: email@example.com\ntoken: some-token\n")
```
rossf7: The tests are great but I'd prefer it the email and token were vars as they are used multiple times.

config/config.go
marians: That's our legacy API stuff. See http://firstgen-docs.giantswarm.io/reference/api/v1/#GetMemberships

We are using that v1 functions still, as it hasn't been migrated to v4 yet. Shame.

config/config.go
marians: Right. Changed.

config/config_test.go
marians: Changed

config/config.go
rossf7: Ah OK I was getting confused between status code and response code. Cool it can be taken care of with the v4 migration.

marians:
@rossf7 Ready for another check

marians:
Thanks for the review!

PR #58, opened 2017-05-12
README update

marians:
- Added `gsctl create cluster` example to usage
- Updated some dates to look a bit fresher
- Changed organization `giantswarm` to `acme`, so it's easier
understood that this is an example value


marians:
Thanks!

Issue #59, opened 2017-05-15
Config migration requires kubeconfig adaptation


marians:
The configuration path change in https://github.com/giantswarm/gsctl/pull/55 requires one more change: The certificate paths in `$HOME/ .kube/config` (or whatever the config file location is) need to be updated, too.

PR #63, opened 2017-05-26
gsctl: Ensure that certificate paths get updated with config dir migration

marians:
Fixes https://github.com/giantswarm/gsctl/issues/59

In #55 a migration of the configuration directory has been introduced. Unfortunately I forgot to take care of a user's kubectl config file.

This PR here makes the migration also maintain the paths. It does it simply by replacing occurrences of the old path with the new path in the config, not trying to be any more clever than that.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -289,14 +289,15 @@ func migrateConfigDir() error {
 	// adapt certificate paths in kubeconfig
 	if len(KubeConfigPaths) > 0 {
 		// we only adapt the first file found (on purpose)
-		if stat, err := os.Stat(KubeConfigPaths[0]); err == nil {
-			oldConfig, configErr := ioutil.ReadFile(KubeConfigPaths[0])
+		theKubeConfigPath := KubeConfigPaths[0]
+		if stat, err := os.Stat(theKubeConfigPath); err == nil {
+			oldConfig, configErr := ioutil.ReadFile(theKubeConfigPath)
 			if configErr != nil {
 				return configErr
 			}
 			newConfig := strings.Replace(string(oldConfig), LegacyConfigDirPath, DefaultConfigDirPath, -1)
 			// write back
-			writeErr := ioutil.WriteFile(KubeConfigPaths[0], []byte(newConfig), stat.Mode())
+			writeErr := ioutil.WriteFile(theKubeConfigPath, []byte(newConfig), stat.Mode())
 			if writeErr != nil {
 				return fmt.Errorf("Could not overwrite kubectl config file. %s", writeErr.Error())
 			}
--- a/config/config.go
+++ b/config/config.go
@@ -299,7 +299,7 @@ func migrateConfigDir() error {
 			// write back
 			writeErr := ioutil.WriteFile(theKubeConfigPath, []byte(newConfig), stat.Mode())
 			if writeErr != nil {
-				return fmt.Errorf("Could not overwrite kubectl config file. %s", writeErr.Error())
+				return errors.New("could not overwrite kubectl config file")
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -92,6 +95,72 @@ func Test_Kubeconfig_Env_Nonexisting(t *testing.T) {
...
+	t.Log("LegacyConfigDirPath:", LegacyConfigDirPath)
+
+	migrateConfigDir()
+
+	// spit out config
+	newConf, confReadErr := ioutil.ReadFile(kubeconfigpath)
+	if confReadErr != nil {
+		t.Error(confReadErr)
+	}
+	t.Log(string(newConf))
```
rossf7: Is this logged to check manually? If so should it be removed?

marians: It will be output if the user starts tests in verbose mode (which btw Atom with go-plus does by default), so I still see a use for it.

config/config.go
```diff
@@ -281,5 +281,27 @@ func migrateConfigDir() error {
...
-	return os.Rename(LegacyConfigDirPath, DefaultConfigDirPath)
+	err = os.Rename(LegacyConfigDirPath, DefaultConfigDirPath)
+	if err != nil {
+		return err
+	}
+
+	// adapt certificate paths in kubeconfig
+	if len(KubeConfigPaths) > 0 {
+		// we only adapt the first file found (on purpose)
+		if stat, err := os.Stat(KubeConfigPaths[0]); err == nil {
```
rossf7: I don't like the `KubeConfigPaths[0]` I'm not sure what this is and makes the code hard to read. I'd also prefer to see a variable as its used multiple times.

config/config_test.go
rossf7: Cool that's fine then. I just wanted to check. BTW I'm currently trying out VS Code and liking it a lot.

config/config.go
marians: Not sure I understood that as you meant it. Changed it. Better?

marians:
@rossf7 I see what you mean, but I couldn't come up with a better alternative. What I'm trying to say is: Of all the possible kubectl configuration paths, this is the f***ing single one I want to deal with here.

config/config.go
```diff
@@ -281,5 +281,28 @@ func migrateConfigDir() error {
...
+		if stat, err := os.Stat(theKubeConfigPath); err == nil {
+			oldConfig, configErr := ioutil.ReadFile(theKubeConfigPath)
+			if configErr != nil {
+				return configErr
+			}
+			newConfig := strings.Replace(string(oldConfig), LegacyConfigDirPath, DefaultConfigDirPath, -1)
+			// write back
+			writeErr := ioutil.WriteFile(theKubeConfigPath, []byte(newConfig), stat.Mode())
+			if writeErr != nil {
+				return fmt.Errorf("Could not overwrite kubectl config file. %s", writeErr.Error())
```
kopiczko: Usually go errors doesn't start with capital letter and terminate with full stop.

marians: True, I'll change that. Tim would also have demanded that. :)

marians:
Thanks for the reviews!

config/config.go
```diff
@@ -299,7 +299,7 @@ func migrateConfigDir() error {
 			// write back
 			writeErr := ioutil.WriteFile(theKubeConfigPath, []byte(newConfig), stat.Mode())
 			if writeErr != nil {
-				return fmt.Errorf("Could not overwrite kubectl config file. %s", writeErr.Error())
+				return errors.New("could not overwrite kubectl config file")
```
kopiczko: I'd say do it like `return fmt.Errorf("could not overwrite kubectl config file: %s", writeErr.Error())`, unless you don't care about writeErr message.

BTW in almost all cases error variable is called `err` and reused.

Issue #56, opened 2017-05-11
Make sure running tests doesn't manipulate kubectl config and add certs


marians:
Currently when running tests, key pairs are added to the certs directory and the user's kubectl is changed, including setting a different context. As a result, the user has more clutter in the certs dir and kubeconfig file and manually has to change the context to work with kubectl.

This should be changed by using a temporary directory and config file.

marians:
Once https://github.com/giantswarm/gsctl/pull/63 is merged, we have an example for creating a temporary kubectl config file, which can be extracted from the one test and used in all of them.

PR #66, opened 2017-05-26
gsctl: Don't overwrite the user's kubectl config when running tests

marians:
Solves https://github.com/giantswarm/gsctl/issues/56

This PR ensures that a temporary kubectl config file is used for test purposes.

Additionally, the output written to the kubectl config file is now verified.

### DIFF
--- a/commands/create_kubeconfig_test.go
+++ b/commands/create_kubeconfig_test.go
@@ -78,4 +78,26 @@ func Test_CreateKubeconfig(t *testing.T) {
 	cmdClusterID = "test-cluster-id"
 	checkCreateKubeconfig(CreateKeypairCommand, []string{})
 	createKubeconfig(CreateKeypairCommand, []string{})
+
+	// check kubeconfig content
+	content, err := ioutil.ReadFile(kubeConfigPath)
+	if err != nil {
+		t.Error(err)
+	}
+
+	// for reference in case of error
+	t.Log(string(content))
+
+	if !strings.Contains(string(content), "current-context: giantswarm-"+cmdClusterID) {
+		t.Error("Kubeconfig doesn't contain the expected current-context value")
+	}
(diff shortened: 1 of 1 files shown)


commands/create_kubeconfig_test.go
```diff
@@ -5,11 +5,48 @@ import (
...
+func tempKubeconfig() (string, error) {
+
+	// override standard paths for testing
+	dir := tempDir()
+	config.HomeDirPath = dir
+	config.DefaultConfigDirPath = path.Join(config.HomeDirPath, ".config", config.ProgramName)
+	config.LegacyConfigDirPath = path.Join(config.HomeDirPath, "."+config.ProgramName)
+
+	// add a test kubectl config file
+	kubeconfigpath := path.Join(dir, "tempkubeconfig")
```
rossf7: Nitpick but the casing of kubeconfigpath is confusing me. Is it so it doesn't conflict with kubeConfigPath?

Sorry I didn't get back to you on the previous PR. I also couldn't think of a better name! Naming is definitely hard and especially on Fridays!

marians: Changed. Just a style inconsistency.

commands/create_kubeconfig_test.go
```diff
@@ -28,6 +65,13 @@ func Test_CreateKubeconfig(t *testing.T) {
...
 	defer mockServer.Close()
 
+	// temporary kubeconfig file
+	kubeConfigPath, err := tempKubeconfig()
+	if err != nil {
+		t.Error(err)
+	}
+	os.Setenv("KUBECONFIG", kubeConfigPath)
+
 	configDir, _ := ioutil.TempDir("", config.ProgramName)
```
kopiczko: You can use `tempDir` here.

Issue #10, opened 2016-12-05
create date parsing is broken when listing clusters


xh3b4sd:
When creating a cluster the time format for the `create_date` field is something like this.

```
2016-12-05T15:41:46.965138231+01:00
```

At the moment the API allows to set the create date explicitly when creating clusters. On Leaseweb we have also these formats. 

```
2016-11-10T10:00:00.000Z
2016-10-19T16:55:26.515752326Z
```

`gsctl` expects the go time format to be this. 

```
2006-01-02T15:04:05.000Z
```

See https://github.com/giantswarm/gsctl/blob/master/util/datetime.go#L15

In case a cluster is now created without providing a value for the `create_date` field, `gsctl` fails to parse the create date when listing clusters. 

```
ID            NAME          CREATED             ORGANIZATION
xh3b4sd-test  xh3b4sd-test  0001 Jan 01, 00:00  test-org
```

We should

1. ensure one definite time format in the cluster service
2. properly parse the create date without assuming some magic format beforehand and apply the format of choice after parsing the date properly

@marians

xh3b4sd:
See https://github.com/giantswarm/cluster-service/issues/32 to track `1.`.

marians:
I'm kind of reluctant to solve this on the client side. We should not have anything but UTC dates/times in our backends.

xh3b4sd:
This does not change the fact that the implementation of time parsing in the client is faulty. The client should simply parse the time string it receives and formats the time to whatever he think is right. Applying magic regular expressions on unknown strings makes no sense as we see. Because this implementation is now broken when the backend is not consistent. Packages should not rely that much on internal knowledge of other systems. That is why we have types for. Using them makes life a little bit more easy in most cases.

marians:
I know that the current implementation isn't cool. Call it a quality trade-off when writing my first Go application.

xh3b4sd:
This is all good. No worries. The issue is about describing a problem which in turn should be fixed at some point.

marians:
TODO:

- Add unit test
- Make parsing more robust to handle different formats

marians:
See also https://github.com/giantswarm/cluster-service/pull/130

marians:
Closed via https://github.com/giantswarm/gsctl/issues/73

PR #73, opened 2017-06-06
Support parsing date format with offset

marians:
See https://github.com/giantswarm/gsctl/issues/10

### DIFF
--- a/util/datetime.go
+++ b/util/datetime.go
@@ -14,12 +14,24 @@ const (
 func ParseDate(dateString string) time.Time {
 	// our standard format
 	template := "2006-01-02T15:04:05.000Z"
+	alternativeTemplate := "2006-01-02T15:04:05.000-07:00"
 
 	// normalizing the number of decimal places to 3
-	re := regexp.MustCompile("\\.[0-9]+Z$")
-	dateString = re.ReplaceAllLiteralString(dateString, ".000Z")
+	// and discarding sub-second detail along the way
+	re := regexp.MustCompile("\\.[0-9]+")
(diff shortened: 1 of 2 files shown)


util/datetime.go
```diff
@@ -6,20 +6,32 @@ import (
...
-	dateString = re.ReplaceAllLiteralString(dateString, ".000Z")
+	// and discarding sub-second detail along the way
+	re := regexp.MustCompile("\\.[0-9]+")
+	dateString = re.ReplaceAllLiteralString(dateString, ".000")
 
-	t, _ := time.Parse(template, dateString)
+	// try parsing with several formats
+	t, err := time.Parse(template, dateString)
+	if err == nil {
+		return t
```
xh3b4sd: Minor nitpick. In case it is not a parse error we might want to return an error to the caller. See https://golang.org/pkg/time/#ParseError. I think at the moment there is only a parse error returned in case there is any error, but this would make it more future proof. Up to you.

marians:
@xh3b4sd You're right, I added https://github.com/giantswarm/cluster-service/pull/130

Since there might still be legacy entries around, I'd still like to make sure gsctl handles these entries gracefully.

util/datetime.go
marians: I kind of like the fact that the functions can be used like `created := util.ShortDate(util.ParseDate(cluster.CreateDate))`. The worst thing that can happen here is that a bad date ( 0000-01-01 ...) gets displayed.

marians:
@xh3b4sd Can this get your approval the way it is?

marians:
Thanks!

PR #74, opened 2017-06-07
On keypair creation, print resulting TTL instead of requested TTL

marians:
Tiny change that lets us see the keypair TTL as received from the backend, as opposed to printing the requested TTL.


marians:
Ping @oponder - It's a very small change. :)

PR #75, opened 2017-06-08
Adding tests for cluster creation

marians:
Part of the long-running quest to have good unit test coverage in gsctl.

### DIFF
--- a/commands/create_cluster_test.go
+++ b/commands/create_cluster_test.go
@@ -1,7 +1,10 @@
 package commands
 
 import (
+	"fmt"
 	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 )
 
@@ -57,7 +60,7 @@ workers:
   - cpu:
       cores: 2
     memory:
-      size_gb: 5
+      size_gb: 5.5
     storage:
       size_gb: 13
     labels:
@@ -73,7 +76,7 @@ workers:
 	if myDef.Workers[1].CPU.Cores != 2 {
 		t.Error("Expected myDef.Workers[1].CPU.Cores to be 2, got: ", myDef.Workers[1].CPU.Cores)
 	}
-	if myDef.Workers[1].Memory.SizeGB != 5.0 {
+	if myDef.Workers[1].Memory.SizeGB != 5.5 {
 		t.Error("Expected myDef.Workers[1].Memory.SizeGB to be 5, got: ", myDef.Workers[1].Memory.SizeGB)
 	}
 	if myDef.Workers[1].Storage.SizeGB != 13.0 {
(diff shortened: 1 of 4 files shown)


commands/create_cluster_test.go
```diff
@@ -56,7 +76,7 @@ workers:
 	if myDef.Workers[1].CPU.Cores != 2 {
 		t.Error("Expected myDef.Workers[1].CPU.Cores to be 2, got: ", myDef.Workers[1].CPU.Cores)
 	}
-	if myDef.Workers[1].Memory.SizeGB != 5.0 {
+	if myDef.Workers[1].Memory.SizeGB != 5.5 {
 		t.Error("Expected myDef.Workers[1].Memory.SizeGB to be 5, got: ", myDef.Workers[1].Memory.SizeGB)
```
rossf7: I think this should be expected 5.5

commands/create_cluster_test.go
```diff
@@ -76,3 +96,47 @@ func Test_CreateFromBadYAML01(t *testing.T) {
...
+
+	// first run, with minimal arguments
+	args := []string{}
+	cmdAPIEndpoint = mockServer.URL
+	cmdOwner = "acme"
+	checkAddCluster(CreateClusterCommand, args)
+	addCluster(CreateClusterCommand, args)
+
+	// second run, with additional arguments
+	cmdClusterName = "UnitTestCluster"
```
rossf7: I find these tests confusing. How are the values getting passed in via the args array? I'd prefer to see a struct defined with the test cases and then loop through them.

marians: I see.

The `args` parameter isn't used at all in the tests. It's just a parameter that's always added when cobra executes the command. I'm not sure, but I think it would be used by cobra if there were additional arguments which aren't mapped to any of the named or positional arguments we have defined.

See https://github.com/spf13/cobra/blob/master/command.go#L90-L92

So basically I'm setting parameters via the `cmd*` variables (like `cmdVerbose`), which are normally set by the user's command line arguments. Then I'm executing the functions  `checkAddCluster` (for validation) and `addCluster`(for final execution).

marians: I'll clean that up, to separate the validation and execution functions from providing output. That should help make the test less confusing.

rossf7: @marians cool thanks that will help. I get know that you're setting the command line vars. I was thinking they'd be set via Cobra or some similar library. 

Thanks for the clarification. I had some gaps in my knowledge of how gsctl works :)

marians:
@rossf7 I'm done here. 😅  Can you give it another look?

marians:
Thanks for reviewing!

As a side effect, the code became much more readable/structured, due to better separation of concerns.

PR #80, opened 2017-06-12
Unified naming for "key pair"

marians:
We used "key-pair" in many places in gsctl, while we use "key pair" in docs and happa. Towards a unified naming!


marians:
Ping @oponder - Should be a simple review. Executed code isn't affected, only printed output and comments.

Issue #79, opened 2017-06-12
"gsctl info" should print effective API endpoint


marians:
When using an alias to set a specific endpoint, the user might not be aware of the currently used endpoint. A quick execution of `myalias info` could help.

So we should print the effective API endpoint URL in the output table.

marians:
Solved

PR #83, opened 2017-06-12
Adding API endpoint output to "gsctl info"

marians:
See #79 

Also refactored functions to separate info acquisition from info printing. That should help keep the code more readable and testable in the longer run.

### DIFF
--- a/commands/info.go
+++ b/commands/info.go
@@ -21,6 +21,34 @@ var (
 	}
 )
 
+// infoArguments represents the arguments we can make use of in this command
+type infoArguments struct {
+	token       string
+	verbose     bool
+	apiEndpoint string
+}
+
+// defaultInfoArguments returns an infoArguments object populated by the user's
+// command line arguments
+func defaultInfoArguments() infoArguments {
+	return infoArguments{
+		token:       cmdToken,
+		verbose:     cmdVerbose,
+		apiEndpoint: cmdAPIEndpoint,
+	}
+}
+
+// infoResult is the struct used to return all the info we might want to print
+type infoResult struct {
+	apiEndpoint     string
+	email           string
+	token           string
+	version         string
+	buildDate       string
+	configFilePath  string
+	kubeConfigPaths []string
+}
+
 func init() {
 	RootCommand.AddCommand(InfoCommand)
 }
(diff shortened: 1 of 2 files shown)


marians:
Preview:

```
API endpoint:         https://foo.bar
Email:                email@example.com
gsctl version:        0.6.0+git
gsctl build:          2017-06-12T20:11:34Z
Config path:          /Users/marian/.config/gsctl/config.yaml
kubectl config path:  /Users/marian/.kube/config
```

commands/info.go
```diff
@@ -21,47 +21,101 @@ var (
...
+
+// defaultInfoArguments returns an infoArguments object populated by the user's
+// command line arguments
+func defaultInfoArguments() infoArguments {
+	return infoArguments{
+		token:       cmdToken,
+		verbose:     cmdVerbose,
+		apiEndpoint: cmdAPIEndpoint,
+	}
+}
```
kopiczko: What is the reason for this thing to exist? It still uses global variables. I don't see tests which inject different values. Is there any plan for this?

commands/info.go
```diff
@@ -21,47 +21,101 @@ var (
...
+
+	fmt.Println(columnize.SimpleFormat(output))
+}
+
+// info gets all the information we'd like to show with the "info" command
+// and returns it as a struct
+func info(args infoArguments) infoResult {
+	result := infoResult{}
+
+	result.apiEndpoint = config.DefaultAPIEndpoint
```
kopiczko: Not sure if mixing global and parameter variables is a good idea in this case. The signature looks like pure function.

commands/info.go
marians: The actual reason is that it simplifies testing. Instead of setting a bunch of global variables between several test cases, this allows to set the arguments specificly for one test case, not affecting other tests.

It's also an alignment with other commands, as I have already started introducing this in previous PRs.

I just haven't rewritten the tests yet.

commands/info.go
marians: Hmm. Don't know what to say really.

`info()` is the function providing all the details that should be printed in the output function for the `gsctl info [-v]` command, as a single struct. It has to pull that info from various places. One of those is the `config` object, which provides a default API endpoint URL.

Issue #72, opened 2017-06-06
"list keypairs" should check if the selected cluster exists


marians:
This output is misleading, as the given cluster ID doesn't exist:

```
gsctl --api-endpoint=https://api-g8s.giantswarm.io list keypairs -c asdf
CREATED  EXPIRES  ID  DESCRIPTION
```

We should inform the user that the selected cluster does not exist (or is not accessible to the user).

See also #71

marians:
Fixed in https://github.com/giantswarm/gsctl/pull/86

Issue #71, opened 2017-06-06
list keypairs should show friendly message if no keypairs exist


marians:
Current implementation:

```
$> gsctl --api-endpoint=https://api-g8s.giantswarm.io list keypairs -c 0xov4
CREATED  EXPIRES  ID  DESCRIPTION
$>
```

We should display a friendly message instead of an empty table. We could also suggets the command to create a new keypair (`create kepair` or `create kubeconfig`).

marians:
This is implemented a little nicer in `list clusters`.

marians:
Fixed in https://github.com/giantswarm/gsctl/pull/86

PR #86, opened 2017-06-14
gsctl list keypairs: Better handling of certain cases

### LINKED ISSUE #64: gsctl list keypairs doesn't note if the user isn't logged in
This happens if the user has a token in the config, but that token isn't valid.

```nohighlight
$ gsctl list keypairs -c asdf
CREATED  EXPIRES  ID  DESCRIPTION
```

The expected output would be

- Tell the user that he's not logged in at the moment for the API endpoint used.

marians:
Fixes #64, #71, #72

### Preview

```nohighlight
$ ./gsctl --api-endpoint=https://api-g8s.giantswarm.io list keypairs -c pmb9q --auth-token="sdfsdf"
You are not authorized for this cluster.
You have no permission to access key pairs for this cluster. Please check your credentials.
```

```nohighlight
$ ./gsctl --api-endpoint=https://api-g8s.giantswarm.io list keypairs -c feogv
No key pairs available for this cluster.
```


marians:
Ping @kopiczko @rossf7

commands/list_keypairs.go
```diff
@@ -23,84 +27,208 @@ var (
...
+// something is missing.
+func listKeypairsValidationOutput(cmd *cobra.Command, extraArgs []string) {
+	args := defaultListKeypairsArguments()
+	err := listKeypairsValidate(&args)
+	if err != nil {
+		var headline string
+		var subtext string
+
+		switch err.Error() {
+		case errNotLoggedIn:
```
kopiczko: Matching error strings is bad.

https://github.com/giantswarm/giantswarm/wiki/go-style-guide#errors

See error matching functions. It would be

```
switch {
case IsNotLoggedIn(err):
```

commands/commands.go
```diff
@@ -45,8 +45,12 @@ var (
...
 
 	// errors for common purposes
-	errNotLoggedIn      = "user not logged in"
-	errConflictingFlags = "conflicting flags used"
+	errNotLoggedIn           = "user not logged in"
+	errConflictingFlags      = "conflicting flags used"
+	errClusterIDNotSpecified = "cluster ID not specified"
+	errClusterNotFound       = "the cluster specified could not be found"
+	errInternalServerError   = "an internal server error occurred"
+	errNotAuthorized         = "not authorized"
```
xh3b4sd: These should be typed errors. Handling errors with strings is an anti pattern in golang. We have types and type assertion. Lets use them.

commands/list_keypairs.go
```diff
@@ -23,84 +27,208 @@ var (
...
+		return errors.New(errNotLoggedIn)
+	}
+	if args.clusterID == "" {
 		// use default cluster if possible
 		clusterID, _ := config.GetDefaultCluster(requestIDHeader, listKeypairsActivityName, cmdLine, cmdAPIEndpoint)
 		if clusterID != "" {
 			cmdClusterID = clusterID
 		} else {
-			return errors.New("No cluster given. Please use the -c/--cluster flag to set a cluster ID.")
+			return errors.New(errClusterIDNotSpecified)
```
xh3b4sd: We should always mask errors to be able to comprehend their stack in case shit hits the fan.

commands/list_keypairs_test.go
```diff
@@ -3,75 +3,146 @@ package commands
...
+	defer os.RemoveAll(dir)
+	config.Initialize(dir)
+
+	// mock service returning key pairs. For the sake of simplicity,
+	// it doesn't care about auth tokens.
 	keyPairsMockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusOK)
 		w.Write([]byte(`[
+			{
```
xh3b4sd: It is only a string but this indentation looks wrong.

commands/list_keypairs.go
kopiczko: I don't know if you mask errors in this project.

commands/list_keypairs.go
```diff
@@ -23,84 +27,208 @@ var (
...
+	// error output
 	if err != nil {
-		fmt.Println(color.RedString("Error: %s", err))
-		if _, ok := err.(APIError); ok {
-			dumpAPIResponse((err).(APIError).APIResponse)
+		var headline string
+		var subtext string
+
+		switch err.Error() {
+		case errNotLoggedIn:
```
kopiczko: Again error matchers.

commands/list_keypairs.go
xh3b4sd: This is rather dangerous. As soon as there is a typo in the error message or it changes the error assertion will fail because we only check for the error string. We should rather use type based error assertions like we do in all of our other projects. Check some microservice and see how it is done.

commands/list_keypairs.go
```diff
@@ -23,84 +27,208 @@ var (
...
+			subtext = "Please notify the Giant Swarm support team, or try listing key pairs again in a few moments."
+		default:
+			headline = err.Error()
+		}
+
+		fmt.Println(color.RedString(headline))
+		if subtext != "" {
+			fmt.Println(subtext)
 		}
 		os.Exit(1)
```
kopiczko: I don't like two things about this project. Using (1) `fmt.Print` directly instead of passing output interface. (2) exiting immediately.
(1) makes testing output difficult.
(2) makes testing difficult and cumbersome, also prevents from using `deffer`. It can be solved by returning errors and exiting in single place.

I know this is change issue, but I think it might be worth tracking that and resolve at some point.

commands/list_keypairs.go
```diff
@@ -23,84 +27,208 @@ var (
...
+		fmt.Println(color.YellowString("No key pairs available for this cluster."))
+		fmt.Println("You can create a new key pair using the 'gsctl create kubeconfig' or 'gsctl create keypair' command.")
+	} else {
+		output := []string{}
+
+		headers := []string{
+			color.CyanString("CREATED"),
+			color.CyanString("EXPIRES"),
+			color.CyanString("ID"),
+			color.CyanString("DESCRIPTION"),
```
kopiczko: I also have mixed filings about using colors directly. Doesn't it make more sense to use functions like `formatHeader`, `formatBody`, etc. ? They can be changed in single place. Also it'd be possible to change them in tests if needed.

commands/list_keypairs.go
```diff
@@ -23,84 +27,208 @@ var (
...
+
 	if err != nil {
-		return "", APIError{err.Error(), *apiResponse}
+
+		if apiResponse.StatusCode >= 500 {
+			return result, errors.New(errInternalServerError)
+		} else if apiResponse.StatusCode == http.StatusNotFound {
+			return result, errors.New(errClusterNotFound)
+		} else if apiResponse.StatusCode == http.StatusUnauthorized {
+			return result, errors.New(errNotAuthorized)
```
kopiczko: Those errors should be defined for matchers. See my other comment.

commands/list_keypairs_test.go
```diff
@@ -3,75 +3,146 @@ package commands
...
+// is logged in or provides an auth token.
+func Test_ListKeypairs_NotLoggedIn(t *testing.T) {
+	defer viper.Reset()
+	dir := tempDir()
+	defer os.RemoveAll(dir)
+	config.Initialize(dir)
 	config.Config.Token = ""
-	err := checkListKeypairs(ListKeypairsCommand, []string{})
-	if err == nil || err.Error() != "You are not logged in. Use 'gsctl login' to log in." {
-		t.Error("Unexpected error:", err)
```
kopiczko: Matchers again.

commands/list_keypairs_test.go
```diff
@@ -3,75 +3,146 @@ package commands
...
-// Not logged in, no cluster ID is given. Should fail.
-func Test_CheckListKeypairs_NotLoggedInNoCluster(t *testing.T) {
-	cmdClusterID = ""
-	err := checkListKeypairs(ListKeypairsCommand, []string{})
-	if err == nil {
-		t.Error("Expected error didn't happen:", err)
+	_, listErr := listKeypairs(args)
+	if listErr == nil {
+		t.Error("No error occurred where we expected one.")
+	} else if listErr.Error() != errClusterNotFound {
```
kopiczko: Matcher.

commands/list_keypairs.go
marians: formatHeader and formatBody would have to be written first. I didn't really intend to go into the depth of `columnize`. At least I don't see a need right now.

marians:
OK guys, thanks for picking on the error stuff. Please keep in mind that this is not a microservice, but a frontend tool.

In my perspective, defining error strings and re-using them was already a step ahead. Since they are defined in one place, I don't see the danger of a string changing in one place and a matcher in another place not matching.

Having said that, I'll see if I can make use of microkit/error in a useful way, hopefully avoiding to add most of microkit's dependencies.

xh3b4sd:
> Never inspect the output of error.Error
>
> As an aside, I believe you should never inspect the output of the error.Error method. The Error method on the error interface exists for humans, not code.
>
> The contents of that string belong in a log file, or displayed on screen. You shouldn’t try to change the behaviour of your program by inspecting it.
>
 > I know that sometimes this isn’t possible, and as someone pointed out on twitter, this advice doesn’t apply to writing tests. Never the less, comparing the string form of an error is, in my opinion, a code smell, and you should try to avoid it.

See https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

commands/list_keypairs_test.go
marians: It's a Github thing. Looks sane in Atom.

commands/list_keypairs.go
marians: I've been changing the code lately to be separate the business logic from the output printing. That at least allows for testing of the business logic, and that works quite well now.

So far we are not testing the printed output in unit tests. Might be that we start doing that at some point.

commands/list_keypairs_test.go
xh3b4sd: It is a spaces vs tabs thing. Here you use tabs. The other code in the string uses spaces. Atom renders it the same way depending on your settings. Github displays spaces and tabs using different widths for this reason, to notice you used different characters. So please change it to spaces.

marians: Ah, I wasn't even aware of that. Thanks! I changed the JSON indentation to tabs.

marians:
@xh3b4sd @kopiczko The error handling is changed to typed errors. I'll have to make that concistent throughout the rest of the code base. Don't want to bring that all into this little PR.

Issue #93, opened 2017-06-20
Scaling up a cluster via gsctl


marians:
We'd like to support adding worker nodes to an existing cluster (a. k. a. scaling up).

Removing worker nodes (scaling down) should be implemented as a next step, too. However, for this spec, removing worker nodes is not in scope.

### Details

- For a cluster with all identical worker nodes, the user should not have to specify any worker details. The command should allow the user to express "I want more nodes like the ones I have".
  - For AWS, in the beginning the "more of what I have" way will be the only way to execute the command, as clusters are restricted to identical worker node types.
- If required, the user may specify details on the new nodes required. This should be aligned with [cluster creation](https://docs.giantswarm.io/reference/gsctl/create-cluster/).
- In cluster creation, we apply a default node configuration if the user doesn't specify node settings. When adding nodes, it should basically work the same way.
- Users can add any number of nodes. Probably a maximum node count limit for the cluster may apply.

### Command Naming

With the current grammar of `gsctl <verb> <noun>`, that function would likely become accessible under a command like

```
gsctl edit cluster
gsctl modify cluster
gsctl change cluster
```

In concordance with `kubectl`, we should probably go with the verb `edit`.

marians:
Question: Do we rather want to let the user express how many nodes to add ("add 2 workers") or how many should exist after scaling ("make it so that 10 nodes exist)? The first isn't idempotent, the last is. But idempotency hasn't been a focus of gsctl commands so far.

marians:
Some drafts how the command could work in an idempotent way:

### Idempotent command draft A

Idempotency here means that the command can be executed several times, as it expresses the desired state.

In this form (long form) the user may specify node details.

1. User executes `gsctl edit cluster <cluster_id>`
2. Default editor opens to edit the cluster definition YAML
3. User adds one or several node entries and saves the file
4. gsctl validates the change, shows error info (optional)
5. If valid, gsctl submits the PATCH request to the API

Expected result: Worker nodes with given settings are added to the cluster, so that the cluster has the specified array of workers.

Validation has to ensure that existing workers are not modified or removed.

### Idempotent command draft B

This form of the command would let the user scale the cluster up to the specified number of nodes, adding nodes with default settings.

Given is a cluster with less than 7 worker nodes.

```
gsctl edit cluster <cluster_id> --num-workers 7
```

In this case, no editor would open and the command would be validated and then executed immediately.

[...]

Issue #94, opened 2017-06-21
delete cluster should use positional argument instead of named argument


marians:
Currently we use this syntax for `gsctl delete cluster`

```
gsctl delete cluster -c <cluster_id>
gsctl delete cluster --cluster <cluster_id>
```

The <cluster_id> argument is technically optional, as there might be a case where the user has only access to one cluster, and then this one woul dbe used as the cluster to delete.

However in practice, most cases require users to specify which cluster to delete. Then it is against intuition that the cluster ID has to be added using a `-c` or `--cluster` named flag.

The syntax should be

```
gsctl delete cluster <cluster_id>
```

Also, in the case of deleting, we should probably step away from supporting the default cluster mechanics described above. The risk of deleting a cluster without explicitly expressing which cluster to delete is probably too high.

PR #106, opened 2017-07-18
Adding scale cluster command

marians:
Towards https://github.com/giantswarm/gsctl/issues/93 https://github.com/giantswarm/giantswarm/issues/1461

### Preview:

```nohighlight
$ gsctl --api-endpoint=http://localhost:9000 scale cluster g1stg -w 5
Desired worker node count equals the current one.
No worker nodes have been added or removed.

$ gsctl --api-endpoint=http://localhost:9000 scale cluster g1stg -w 6
The cluster is being scaled up
Adding 1 worker to the cluster for a total of 6 workers.

$ gsctl --api-endpoint=http://localhost:9000 scale cluster g1stg -w 5
Do you really want to reduce the worker nodes for cluster 'g1stg' to 5? [y/n]: y
The cluster is being scaled down
Removing 1 worker from the cluster for a total of 5 workers.

$ gsctl scale cluster --help
Increase or reduce the number of worker nodes in a cluster.

Caution:

When reducing the number of nodes, the selection of the node(s) to be removed
is non-deterministic. Workloads on the worker nodes to be removed will be
terminated, data stored on the worker nodes will be lost. Make sure to remove
only as many nodes as your deployment architecture can handle in a resilient
way.

Examples:

  gsctl scale cluster c7t2o -w 12

  gsctl scale cluster c7t2o -w 12 --num-cpus 4

Usage:
  gsctl scale cluster [flags]

Flags:
      --force                If set, no confirmation is required when reducing the number of workers.
  -h, --help                 help for cluster
      --memory-gb float32    RAM per added worker node.
      --num-cpus int         Number of CPU cores per added worker node.
  -w, --num-workers int      Number of worker nodes to have after scaling.
      --storage-gb float32   Local storage size per added worker node.

Global Flags:
      --api-endpoint string   The URL base path, to access the API (default "https://api.giantswarm.io")
      --auth-token string     Authorization token to use
      --config-dir string     Configuration directory path to use (default "/Users/marian/.config/gsctl")
  -v, --verbose               Print more information
```

### TODO

- [x] Allow for setting config details of workers to be added (KVM only)
- [x] Test against working API
- [x] Change gsclientgen branch name to master once https://github.com/giantswarm/gsclientgen/pull/15 is merged


calvix:
I find it weird that when deleting cluster I have to specify via `-c` and when scaling I don't have to do that.
I would vote for consistency

```
gsctl scale cluster -c c7t2o -w 12
```

marians:
@calvix Me too. That will be fixed. https://github.com/giantswarm/gsctl/issues/94

marians:
Just noticed that a unit test needs fixing. On it.

calvix:
```
      --memory-gb float32    RAM per added worker node.
      --num-cpus int         Number of CPU cores per added worker node.
      --storage-gb float32   Local storage size per added worker node.
```
This is intresting, This look super complex from the operator size.
So if i scale from 4 workers to 8 and also increase CPU from 2 to 6.
This means we may have to reconfigure autoscaling group or maybe even create a new one? Hard to say but this may lead into serious downtime if handled badly.

Also we should distinguish between aws and kvm, becasue in aws you cant specify cpu or ram, you need to specify instance type.

marians:
These flags have no effect on AWS.

See here for more details on how scaling works and what restrictions apply: https://github.com/giantswarm/api-spec/blob/scale-cluster-2/spec.yaml#L239-L349

marians:
Unit test is fixed.

commands/scale_cluster.go
```diff
@@ -0,0 +1,348 @@
...
+		PreRun: scaleClusterPreRunOutput,
+
+		// Run calls the business function and prints results and errors.
+		Run: scaleClusterRunOutput,
+	}
+)
+
+const (
+	scaleClusterActivityName string = "scale-cluster"
+	getClusterActivityName   string = "get-cluster"
```
kopiczko: nitpick: no need to specify `string` type here.

commands/scale_cluster.go
```diff
@@ -0,0 +1,348 @@
...
+	if result.numWorkersToAdd > 0 {
+		// scaling up
+		fmt.Println(color.GreenString("The cluster is being scaled up"))
+		if result.numWorkersToAdd == 1 {
+			fmt.Printf("Adding 1 worker to the cluster for a total of %d workers.\n",
+				result.numWorkersAfter)
+		} else {
+			fmt.Printf("Adding %d workers to the cluster for a total of %d workers.\n",
+				result.numWorkersToAdd, result.numWorkersAfter)
+		}
```
kopiczko: I'd do it like

```
workerWord := "workers"
if result.numWorkersToAdd > 1 {
        workerWord = "worker"
}
fmt.Printf("Adding %d %s to the cluster for a total of %d workers.\n",
        result.numWorkersToAdd, workerWord, result.numWorkersAfter)
```

Or even create a closure:

```
workerWord := func(i int) string {
        if i == 1 {
                return "worker"
        }
        return "workers"
}

marians: Good inspiration!

marians:
@kopiczko Since you reviewed this, I changed the API call logic a bit, to remove a superfluous GET call for cluster details after scaling. Also adapted the test again.

Issue #67, opened 2017-05-29
automatically check for a newer gsctl 


marians:
We should automatically check for a newer gsctl version every now and then (like once every 24h).

Story: _As a user, I want to know when a newer version of gsctl is available, so I can benefit from feature additions and bugfixes_

We have to make sure that this doesn't interfere with command calls which expect any interactive input, like `login` or `delete cluster`. In the case of these commands, the new version info may be displayed after a successful transaction.

The latest version can be obtained via this URL:

`http[s]://downloads.giantswarm.io/gsctl/VERSION`

The time of the last version check should be stored in the config file.

If a newer version is found, we can display something like this:

#### Mac OS/brew

```nohighlight
Good news: the newer gsctl version 0.6.0 is now available for download. To install:

    brew upgrade gsctl

Release notes: https://github.com/giantswarm/gsctl/releases/tag/0.6.0
```

#### Mac non-brew

```nohighlight
Good news: the newer gsctl version 0.6.0 is now available for download. To install:

    curl -O http://downloads.giantswarm.io/gsctl/0.6.0/gsctl-0.6.0-darwin-amd64.tar.gz
    tar xzf gsctl-0.6.0-darwin-amd64.tar.gz
    sudo cp gsctl-0.6.0-darwin-amd64/gsctl /usr/local/bin/
    rm -r gsctl-0.6.0-darwin-amd64

Release notes: https://github.com/giantswarm/gsctl/releases/tag/0.6.0
```

#### Linux

```nohighlight
Good news: the newer gsctl version 0.6.0 is now available for download. To install:

    curl -O http://downloads.giantswarm.io/gsctl/0.6.0/gsctl-0.6.0-linux-amd64.tar.gz
    tar xzf gsctl-0.6.0-linux-amd64.tar.gz
    sudo cp gsctl-0.6.0-linux-amd64/gsctl /usr/local/bin/
    rm -r gsctl-0.6.0-linux-amd64

Release notes: https://github.com/giantswarm/gsctl/releases/tag/0.6.0
```

#### Windows

```nohighlight
Good news: the newer gsctl version 0.6.0 is now available for download. To install, download and
extract the right ZIP for your platform:

    64 bit: http://downloads.giantswarm.io/gsctl/0.6.0/gsctl-0.6.0-windows-amd64.zip
    32 bit: http://downloads.giantswarm.io/gsctl/0.6.0/gsctl-0.6.0-windows-386.zip

[...]

PR #108, opened 2017-07-20
Adding update check (to version command)

marians:
For https://github.com/giantswarm/gsctl/issues/67

Preview:

image


marians:
@oponder @rossf7 Are you able to review this, so it can get into today's release?

commands/version.go
```diff
@@ -26,16 +40,103 @@ func init() {
...
+			fmt.Println()
+			fmt.Println(formatUpdateInfo(info))
 		}
 	}
 }
+
+// latestVersion returns the latest available version as semver.Version
+func latestVersion(url string) (*semver.Version, error) {
+	// to be unobstructive, we timeout quickly.
+	timeout := time.Duration(2 * time.Second)
```
rossf7: It would be nice if this was a constant rather than inline.

marians:
The build error occurred only on CircleCI. We normally use Travis here in gsctl. I turned Circle off again, since I couldn't fix the problem yet. But that's unrelated to this PR.

Issue #110, opened 2017-07-28
gsctl fails to run on alpine - user.Current() not implemented on linux/amd64


marians:
### What did you do/observe?

```
$ gsctl --help
Could not get system user details using os/user.Current().
Without this information, gsctl cannot determine the user’s home directory and cannot set a configuration path.
Please get in touch with us via support@giantswarm.io, including information on your platform.
Thank you and sorry for the inconvenience!

panic: user: Current not implemented on linux/amd64

goroutine 1 [running]:
github.com/giantswarm/gsctl/config.init.1()
    /go/src/github.com/giantswarm/gsctl/config/config.go:93 +0x5bc
github.com/giantswarm/gsctl/config.init()
    /go/src/github.com/giantswarm/gsctl/config/config.go:302 +0x7b
github.com/giantswarm/gsctl/commands.init()
    /go/src/github.com/giantswarm/gsctl/commands/version.go:42 +0x9b
main.init()
    /go/src/github.com/giantswarm/gsctl/main.go:26 +0x58
```

### Details

- Operating system, platform/architecture:

```
$ uname -a
Linux 6dd830902cce 4.9.36-moby #1 SMP Wed Jul 12 15:29:07 UTC 2017 x86_64 Linux
```

- Output of `gsctl version`: 0.6.3

- Any other details required to reproduce the problem:
  - Inside a Docker container.

marians:
This also happens in a debian based Docker container.

```
$ gsctl
Could not get system user details using os/user.Current().
Without this information, gsctl cannot determine the user’s home directory and cannot set a configuration path.
Please get in touch with us via support@giantswarm.io, including information on your platform.
Thank you and sorry for the inconvenience!

panic: user: Current not implemented on linux/amd64

goroutine 1 [running]:
github.com/giantswarm/gsctl/config.init.1()
    /go/src/github.com/giantswarm/gsctl/config/config.go:93 +0x5bc
github.com/giantswarm/gsctl/config.init()
    /go/src/github.com/giantswarm/gsctl/config/config.go:302 +0x7b
github.com/giantswarm/gsctl/commands.init()
    /go/src/github.com/giantswarm/gsctl/commands/version.go:42 +0x9b
main.init()
    /go/src/github.com/giantswarm/gsctl/main.go:26 +0x58

$ uname -a
Linux 0613f7834bbb 4.9.36-moby #1 SMP Wed Jul 12 15:29:07 UTC 2017 x86_64 GNU/Linux

[...]

PR #111, opened 2017-07-28
Make gsctl work in Docker containers

marians:
Attempt to fix https://github.com/giantswarm/gsctl/issues/110

Status: work in progress

This PR enables using `gsctl` in a Docker container.

To do that, it enables CGO (`CGO_ENABLED=1`) for C bindings, only for the linux binary.

As a consequence the binary is no longer statically linked. This means it requires glibc around. So it won't work in a plain `alpine` Docker image.

Binary for test purposes: [gsctl.zip](https://github.com/giantswarm/gsctl/files/1183432/gsctl.zip)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PROJECT=gsctl
 ORGANISATION=giantswarm
 BIN=$(PROJECT)
-GOVERSION := 1.8-alpine
+GOVERSION := 1.8
 BUILDDATE := $(shell date -u +%Y-%m-%dT%H:%M:%SZ)
 VERSION := $(shell cat VERSION)
 COMMIT := $(shell git rev-parse HEAD | cut -c1-10)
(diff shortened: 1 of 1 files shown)


marians:
@hobti01 The binary in the ZIP linked above works in a Docker container based on `debian:latest`. I added nothing bot the `gsctl` binary. It's not expected to work with `alpine`, as that doesn't contain glibc.

If alpine support is required, we can work on producing an extra build.

There seems to be a workaround, which I haven't tested so far:
https://github.com/sgerrand/alpine-pkg-glibc

marians:
I tested https://github.com/sgerrand/alpine-pkg-glibc now and it works with the gsctl binary above.

Here is my Dockerfile

```dockerfile
FROM alpine:3.6

RUN apk update && apk --no-cache add ca-certificates openssl && \
  update-ca-certificates && \
  wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub && \
  wget -q https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.25-r0/glibc-2.25-r0.apk && \
  apk add glibc-2.25-r0.apk

COPY ./gsctl /

CMD /gsctl
```

marians:
If possible, gsctl should work in a variety of different Docker containers using different base images. Reason: users might have different additional requirements for the containers, since they usually do other things in a CI/CD-pipeline as well, not only execute gsctl.

So here we have ways to make it work with both Ubuntu (glibc) and Alpine (musl-libc). Trying to find out if this is enough for now.

hobti01:
@marians this is great!  Not ideal that glibc must be added to alpine, but not a blocker.

marians:
@kopiczko @xh3b4sd @JosephSalisbury @rossf7 Anyone for review/lgtm?

xh3b4sd:
I think when this does the do for you now lets move on. When looking at the changes and how the build is done I wonder why we don't want to have all the magic in `architect` and build all CLI tools the same way. Just food for thought.

xh3b4sd:
Oh, one update. The CI build is failing. So lets get the lights green before merging this.

marians:
This is an annoying one. I added circle.yml in a different branch, which isn't even merged yet. Not Circle thinks it can doe CI here, but in fact it can't. And that's not a simple fix. But I could turn off circle for this repo.

marians:
@xh3b4sd Checks are green. I turned off Circle.

marians:
Can somebody please give this an approval? @kopiczko @xh3b4sd

marians:
Thanks!

PR #115, opened 2017-09-01
Add cn_prefix and certificate_organizations flags to create keypair command

oponder:


oponder:
ping @puja108, just in case this got lost over the weekend

oponder:
No docs update yet, I'll make that now.

Also made me realise it would be handy to have these params also on the `create kubeconfig` command, so I'll add that to this PR.

puja108:
oh, yes, would be good to have this in create kubeconfig

PR #116, opened 2017-09-05
Force up to date golang.org/x/sys/windows since another package was requesting an out of date version

oponder:
I wasn't able to compile gsctl for windows due to a dependency error.
Turns out one package was forcing an out of date dependency that another package needed the
most recent changes for.


oponder:
Travis failed:
```
0.35s$ sudo apt-get install -y glide
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package glide
```

investigating....

oponder:
Just some temporary glitch in the matrix.

Issue #47, opened 2017-04-20
gsctl should handle multiple API endpoints as contexts


marians:
Story: As a user, I want to stay logged in while using more than one API endpoint, so that I can quickly switch between installations (contexts).

Currently a config file stores only one `email` and `token`. This forces users to log in every time when switching to a different API endpoint using the `--api-endpoint` flag.

We could instead store the endpoint-specific context information like this:

```yaml
endpoints:
  "https://api.giantswarm.io":
    token: ecb96eeb-3e16-4c04-8673-fb600bc36882
    email: marian@giantswarm.io
    organization: ""
    cluster: ""
updated: 2017-04-20T10:28:08+02:00
```

This would allow to easily keep login data for any number of endpoints.

marians:
Note: With a change as described above, we'll have to introduce explicit context switching. Similar to `kubectl config set-context`, but hopefully easier to use (less typing, autocomplete, ...?)

r7vme:
Can also be realted here: https://github.com/giantswarm/gsctl/issues/104

marians:
Some spec as request for comments:

### Concept of endpoint switching

The "endpoint" is the API endpoint URL currently configured via the `--api-endpoint` option on every command.

Endpoint switching allows to persistently set the endpoint for all subsequent commands. The currently selected endpoint can be overridden explicitly (using a command line option).

The configuration file automatically stores each endpoint used. The token per is stored per endpoint, so that users can stay logged in with different endpoints.

As of now, gsctl assumes `https://api.giantswarm.io` as the default endpoint. That didn't prove to be useful, as most users are on installations using other endpoints, so there won't be a default endpoint any longer.

If no endpoint is configured, which will be the case on first usage, and a command is executed without `--endpoint` option, an error message will hunch the user to

- either use `gsctl login --endpoint` to add an endpoint
- or use the `--endpoint` and `--auth-token` options with the given command

### Adding an endpoint

Currently users use `gsctl login --api-endpoint ...` in order to start a session with a specific endpoint.

This will continue to work. However the `--api-endpoint` option will be deprecated. It will be replaced by `--endpoint` for consistency reasons. So the canonical way to add an endpoint will be:

```
gsctl login --endpoint https://api.mydomain.com user@mydomain.com
```

As `https://` is the only scheme we support, that port should become optional so that the above command could also be entered as `gsctl login --endpoint api.mydomain.com ...`.

The login command automatically selects the endpoint for further commands. Read on for details.

[...]

PR #120, opened 2017-09-28
Endpoint switching

### LINKED ISSUE #92: gsctl commands ignore --auth-token given via flag
This should actually trigger a "not authenticated" error:
```
gsctl --api-endpoint=https://api-g8s.giantswarm.io list clusters --auth-token=sfdsdf
```

But it doesn't.

marians:
Implementation of #47

Fixes #92 along the way (making use of `--auth-token`)

### Information for testers

When released, this sort if information will be communicated to users via the CHANGELOG and also in the docs and README.

##### Config is partly incompatible with previous versions

The configuration file (by default: `$HOME/.config/gsctl/config.yaml`) will be overwritten with per-endpoint auth data. Switching between the a released version and this version will mean that you'll have to re-login and endpoint info gets lost.

##### Logging in, adding and selecting endpoint:

```nohighlight
gsctl login <email> -e <endpoint>
```

##### Listing all endpoints:

```nohighlight
gsctl list endpoints
```

##### Switching to a different endpoint:

```nohighlight
gsctl select endpoint <endpoint>
```

Alternatively, the environment variable `GSCTL_ENDPOINT` can be defined or the flag `-e`/`--endpoint` can be used with every command. Flag overwrites variable. Variable overwrites selected endpoint.

##### Check what's the current endpoint:

Two options:

```nohighlight
gsctl info
gsctl list endpoints
```

### Changes

- Removed default API endpoint logic
- Added global `-e` / `--endpoint` flag
- Marked `--api-endpoint` flag as deprecated
- Removed `microkit` dependency, replaced with `microerror`
- Removed dependency `github.com/spf13/viper`, as it doesn't help us with our nested config file
- `gsctl info` now also prints whether or not the user is logged in
- Config will now keep the email address of a user after logout. This enables re-login without having to add the email again to `gsctl login`.
- `config.Initialize()` now resets the config struct, to make test cases work.
- `gsctl logout` now gives output.
- If the endpoint is set via a flag (`-e`/`--endpoint`) the auth logic is as follows: If a token is given via command line flag, that one is used. Otherwise, if a token is stored for the endpoint, that one is used. Otherwise we don't have a token.

### TODO

- [x] Let `gsctl login` add new endpoint if unknown endpoint URL is used
- [x] Normalize API endpoint URL entered by user as `-e` or `--endpoint` flag. Add `https://`, remove trailing slash, remove default port 443 for HTTPS or 80 for HTTP.
- [x] Let `gsctl list endpoints` print all endpoint URLs from config
- [x] Let `gsctl select endpoint ...` switch endpoints
- [x] When `GSCTL_ENDPOINT=foo ./gsctl list endpoints` is executed, non of the listed endpoints must be selected.
- [x] Implement evaluation of environment variable `GSCTL_ENDPOINT`
- [x] Gracefully handle execution of `gsctl login` without an endpoint flag
- [x] For every command, make sure endpoint selection order is flag > env variable > selected endpoint
- [x] The selection which auth token is used should depend on the endpoint selection logic. If the endpoint is given via `-e` flag or coming from the env variable, the token should come from a flag (or from a variable). The selected endpoint's token may only be used if it's the same as given via flag/variable.


oponder:
When using the `-e` flag, what auth token gets used? It seems to use the auth token of the currently selected endpoint. This will only work if the endpoint the user specifies with -e is the same as the currently selected endpoint.

In all other cases, this will probably never lead to a satisfying result.

Suggestion:

- Use the auth token for the given endpoint if it exists in the config
- And, complain that the user has never logged in with a given endpoint if there is no entry for it in the config.
- Don't complain or try to use any stored auth token if the user also provides the auth token flag.

marians:
@oponder Commit https://github.com/giantswarm/gsctl/pull/120/commits/04fe700c9e597f1c2d3627c84d85f474e4329271 brings the change you requested. The behaviour of `config.ChooseToken`:

```
ChooseToken chooses a token to use, according to a rule set.
- If the given token is not empty, we use (return) that
- If the given token is empty and we have an auth token for the given
  endpoint, we return that
- otherwise we return an empty string
```

oponder:
WFM!

client/error.go
```diff
@@ -0,0 +1,11 @@
+package client
+
+import "github.com/juju/errgo"
+
+// attempt to create a client without endpoint
+var endpointNotSpecifiedError = errgo.New("no endpoint has been specified")
```
kopiczko: Godoc usually starts with the variable/member/function name.

client/client.go
```diff
@@ -37,6 +37,11 @@ type GenericResponse struct {
 // with specific configuration
 func NewClient(clientConfig Configuration) (*gsclientgen.DefaultApi, error) {
 	configuration := gsclientgen.NewConfiguration()
+
+	if clientConfig.Endpoint == "" {
+		return &gsclientgen.DefaultApi{}, microerror.Mask(endpointNotSpecifiedError)
+	}
```
kopiczko: We ususally use

```
var invalidConfigError = microerror.New("invalid config")

// IsInvalidConfig asserts invalidConfigError.
func IsInvalidConfig(err error) bool {
	return microerror.Cause(err) == invalidConfigError
}
```
And then

```
return nil, microerror.Maskf(invalidConfigError, "endpoint must not be empty")
```

commands/commands_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/giantswarm/gsctl/config"
+)
+
+// create a temporary directory
+func tempDir() string {
+	dir, _ := ioutil.TempDir("", config.ProgramName)
```
kopiczko: I know it's out of scope, but: I don't think ignoring this error is a good idea. You can either pass `t  *testing.T` and call `t.Fail` or `panic`.

commands/list_endpoints.go
```diff
@@ -0,0 +1,68 @@
...
+
+// endpointsTable returns a table of clusters the user has access to
+func endpointsTable() string {
+	if len(config.Config.Endpoints) == 0 {
+		return fmt.Sprintf("No endpoints configured.\n\nTo add an endpoint and authenticate for it, use\n\n\t%s\n",
+			color.YellowString("gsctl login <email> -e <endpoint>"))
+	}
+
+	// table headers
+	output := []string{color.CyanString("ENDPOINT URL") + "|" + color.CyanString("EMAIL") + "|" + color.CyanString("SELECTED") + "|" + color.CyanString("LOGGED IN")}
```
kopiczko: You can use `strings.Join` to shorten those lines. https://golang.org/pkg/strings/#Join

commands/list_endpoints_test.go
```diff
@@ -0,0 +1,31 @@
...
+	dir, err := tempConfig(yamlText)
+	if err != nil {
+		t.Error(err)
+	}
+	defer os.RemoveAll(dir)
+
+	table := endpointsTable()
+	if table == "" {
+		t.Error("Got no output where I expected a table")
+	}
```
kopiczko: Why not test against actual output?

commands/list_keypairs_test.go
```diff
@@ -5,22 +5,20 @@ import (
...
+		t.Error(err)
+	}
 	defer os.RemoveAll(dir)
-	config.Initialize(dir)
-	config.Config.Token = ""
 
 	args := listKeypairsArguments{}
-	err := listKeypairsValidate(&args)
+	err = listKeypairsValidate(&args)
 	if err == nil {
```
kopiczko: You can use matcher function to check if this is the desired error.

commands/logout.go
```diff
@@ -42,33 +40,30 @@ type logoutArguments struct {
...
 	RootCommand.AddCommand(LogoutCommand)
 }
 
-// TODO: separate validation and validation result output
-func logoutValidationOutput(cmd *cobra.Command, args []string) error {
+func logoutValidationOutput(cmd *cobra.Command, args []string) {
 	if config.Config.Token == "" && cmdToken == "" {
-		return errors.New("You are not logged in")
+		fmt.Println("You weren't logged in here, but better be safe than sorry.")
+		os.Exit(1)
```
kopiczko: Why is it better to exit immediately than return error?

config/config.go
```diff
@@ -75,29 +69,141 @@ var (
...
+
+	// Email is the user email found for the selected endpoint. Might be empty.
+	// Not marshalled back to the config file, as it is contained in the
+	// endpoint's entry.
+	Email string `yaml:"-"`
+}
+
+// endpointConfigStruct is used to serialize/deserialize endpoint configuration
+// to/from a config file
+type endpointConfigStruct struct {
```
kopiczko: Having `Struct` suffix in a struct name is a little bit awkward.

config/config.go
```diff
@@ -75,29 +69,141 @@ var (
...
+func (c *configStruct) Logout(endpointURL string) {
+	ep := normalizeEndpoint(endpointURL)
+
+	if ep == c.SelectedEndpoint {
+		c.Token = ""
+	}
+
+	if element, ok := c.Endpoints[ep]; ok {
+		element.Token = ""
+	}
```
kopiczko: Why not `delete(c.Endpoints, ep)`?

config/config.go
```diff
@@ -120,23 +225,19 @@ func init() {
 // It's supposed to be called after init().
 // The configDirPath argument can be given to override the DefaultConfigDirPath.
 func Initialize(configDirPath string) error {
+	// Reset our Config object. This is particularly necessary for running
+	// multiple tests in a row.
+	Config = configStruct{}
```
kopiczko: I really dislike this is a global variable. This also makes `WriteToFile` very magical.

client/error.go
marians: I do that for exported variables (uppercase first letter), but not for non exported ones. IMHO the dublication makes the comment harder to read directly in the code.

client/client.go
marians: Cool. Learnt something new. That improves things. Can we keep that to a new PR? I basically tried to swap microkit with microerrer (which I understand was a mistake in hindsight)

commands/commands_test.go
marians: panic makes sense. It's used in tests only, so a non-graceful exit would be fine.

client/error.go
kopiczko: > IMHO the duplication makes the comment harder to read directly in the code.

I understand it doesn't fit your personal taste. This is established convention, so let's conform.

commands/list_endpoints.go
marians: Very handy!

commands/logout.go
marians: Because this we way can influence the error output shown to the user. If we would return an error here, cobra would show the error text and the entire command usage.

config/config.go
marians: Yeah. Changed.

config/config.go
marians: I want to keep the endpoint entry with the email address stored (`element.Email`), to simplify future logins.

config/config.go
marians: What would be a better way?

kopiczko: I checked out the code.
What is confusing that there are methods of the config struct itself, functions (i.e. `WriteToFile`) which rely on the fact that `config.Config` is a global variable.

I'd be ok for me if either `WriteToFile` took a config argument or config struct had method `WriteToFile`.

marians:
@kopiczko Does this get your approval?

PR #135, opened 2017-11-08
Fix releases

### LINKED ISSUE #131: gsctl release process needs fixing, is broken due to builder
Last time I released a gsctl version I ran into problems with builder not creating the tag/release as it used to do.

In addition the upload to AWS S3 needs credentials as environment variables. This should be checked early and the user should get a proper warning.

marians:
Fixes #131 (mostly)

- Changed from using builder to manual tagging
- Improved release documentation, especially with respect to `aws` and MFA

Binary upload to github so far doesn't work. Don't know why. But that hasn't been the case before. Binaries must be added to the release draft manually.


marians:
It would be nice, but I don't really see how we could create releases with proper semver versioning automatically.

Issue #128, opened 2017-10-18
Prevent timeouts for "gsctl list clusters"


marians:
### What did you do/observe?

`gstl list clusters` yielded a timeout error 5 times in a row on an AWS installation. After the 6th try, the 3 available clusters got listed.

### What did you expect?

`gsctl list clusters` should show a list of clusters after one or two seconds at most.

### Details

The installation has many organizations. Does this mean we are using many requests to list all the clusters?

PR #136, opened 2017-11-09
Tweak to prevent timeouts in 'list clusters'

marians:
Fixes https://github.com/giantswarm/gsctl/issues/128

This increases the timeout for the API request issues by `gsctl list clusters`. Seems like that request can take a while when the user is a member of many organizations.


marians:
@oponder It's really only a very tiny change to review

Issue #95, opened 2017-06-21
create cluster: AWS instance type should be selectable via command line


marians:
When creating a new cluster, it should be possible to use a flag like `--aws-instance-type` to set the EC2 instance type. Currently this is only possible when submitting a YAML file.

PR #138, opened 2017-11-09
Select instance type when creating a cluster

marians:
Fixes https://github.com/giantswarm/gsctl/issues/95

Depends on https://github.com/giantswarm/gsclientgen/pull/17

### TODO

- [x] merge gsclientgen into master, change branch name to master here in glide.yaml and update dependencies.


commands/error.go
```diff
@@ -206,6 +206,22 @@ func IsCannotScaleBelowMinimumWorkersError(err error) bool {
 	return errgo.Cause(err) == cannotScaleBelowMinimumWorkersError
 }
 
+// user has selected an unknown AWS EC2 instance type
+var invalidAwsEc2InstanceTypeError = errgo.New("invalid AWS EC2 instance type selected")
```
oponder: I see this new error, but I don't see it actually being used anywhere. Is there any validation on the instance type string? (besides conflicting flag validation that I see in this PR)

commands/create_cluster.go
```diff
@@ -245,11 +250,11 @@ func validateCreateClusterPreConditions(args addClusterArguments) error {
 
 	// false flag combination?
 	if args.inputYAMLFile != "" {
-		if args.numWorkers != 0 || args.workerNumCPUs != 0 || args.workerMemorySizeGB != 0 || args.workerStorageSizeGB != 0 {
+		if args.numWorkers != 0 || args.workerNumCPUs != 0 || args.workerMemorySizeGB != 0 || args.workerStorageSizeGB != 0 || args.wokerAwsEc2InstanceType != "" {
 			return microerror.Mask(conflictingFlagsError)
 		}
 	} else {
-		if args.numWorkers == 0 && (args.workerNumCPUs != 0 || args.workerMemorySizeGB != 0 || args.workerStorageSizeGB != 0) {
+		if args.numWorkers == 0 && (args.workerNumCPUs != 0 || args.workerMemorySizeGB != 0 || args.workerStorageSizeGB != 0 || args.wokerAwsEc2InstanceType != "") {
```
oponder: I'm a bit suspicious about this line. have you tested `gsctl` already with this new flag? I haven't taken the time to build it and try it yet. But this makes me think we'll always get a 'numWorkerNodesMissingError' when using the new instance type flag. Or does `args.numWorkers` get set by default somewhere before these preconditions?

commands/error.go
marians: Good catch! That's a remainder from when i had validation implemented.

commands/create_cluster.go
marians: I have tested, yes. In fact the `--aws-instane-type` can only be used if also the `--num-workers` flag is set. Otherwise we can't generate an appropriate cluster definition to submit. We need `n` worker items, and each holds the aws.instance_type attribute. Same is the case for the CPU, RAM, and storage flags.

commands/error.go
oponder: There is server side validation luckily for this value, so I'm fine if gsctl doesn't validate :)

commands/create_cluster.go
oponder: So the default behaviour of 3 nodes is not a thing when we're doing aws-instance-type?

marians:
I removed the unused error. Travis check has passed, Circle hasn't. I just re-enabled Circle because I want to finish up the switch to circle, but untuil that's done the circle tests can't pass.

commands/create_cluster.go
marians: Right. With the way our API is designed, we can't express "I want the default amount of workers, and they have to be this non-default instance type".

PR #154, opened 2017-11-24
Dependency update

marians:
### TODO

- [ ] switch back to `master` once https://github.com/giantswarm/gsclientgen/pull/21 is merged

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 41f48a14ab8eba9db1b0284e66aabb9ef340edfb220edbbfd195e0124e513845
-updated: 2017-11-21T15:40:52.825574+01:00
+hash: de163f7da8d8ed1686a5b801367cea706d01eb9c9d58525167f430502ab6a58f
+updated: 2017-11-24T11:56:29.096696+01:00
 imports:
 - name: github.com/bradfitz/slice
   version: d9036e2120b5ddfa53f3ebccd618c4af275f47da
@@ -14,7 +14,7 @@ imports:
 - name: github.com/giantswarm/columnize
   version: 677ded9e47e75ca49bd2541794a7281a49a2f17b
 - name: github.com/giantswarm/gsclientgen
-  version: 9a965a88d40f2be1601415664d31ea9ab258f7ea
+  version: 2dd9a6c9ab5dafb41d93f389fb39befe1ff6b4c0
 - name: github.com/giantswarm/microerror
   version: ec5f66bb657662031c0d651504fee6908cdf28d9
 - name: github.com/hashicorp/go-rootcerts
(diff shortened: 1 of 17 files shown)


commands/create_cluster.go
```diff
@@ -111,30 +113,39 @@ Examples:
...
+	// cmdRelease sets a release to use, provided as a command line flag
+	cmdRelease string
 	// dry run command line flag
 	cmdDryRun bool
 )
 
 func init() {
 	CreateClusterCommand.Flags().StringVarP(&cmdInputYAMLFile, "file", "f", "", "Path to a cluster definition YAML file")
-	CreateClusterCommand.Flags().StringVarP(&cmdClusterName, "name", "", "", "Cluster name")
+	CreateClusterCommand.Flags().StringVarP(&cmdClusterName, "name", "n", "", "Cluster name")
```
kopiczko: Ninja? :D

commands/create_cluster.go
```diff
@@ -222,6 +237,12 @@ func createClusterExecutionOutput(cmd *cobra.Command, args []string) {
 			headline = "The cluster could not be created."
 			subtext = "You might try again in a few moments. If that doesn't work, please contact the Giant Swarm support team."
 			subtext += " Sorry for the inconvenience!"
+
+			if richErrorOK {
+				subtext += "\n\nDetails:\n"
+				subtext += richError.Message()
+			}
```
kopiczko: Isn't it better to add `else` and print any error even if the output is ugly? How user can create an issue is something isn't working?

commands/create_cluster_test.go
```diff
@@ -131,13 +154,12 @@ func Test_CreateClusterSuccessfully(t *testing.T) {
...
-			token:             "fake token",
-			inputYAMLFile:     "testdata/minimal.yaml",
-			kubernetesVersion: "K8sVersionFromArgs",
-			verbose:           true,
+			apiEndpoint:   mockServer.URL,
+			clusterName:   "Cluster Name from Args",
+			owner:         "acme",
+			token:         "fake token",
+			inputYAMLFile: "testdata/minimal.yaml",
+			verbose:       true,
```
kopiczko: No releaseVersion here?

commands/create_cluster.go
marians: You mean, even if the error was not of type `errgo.Err`, print these details?

commands/create_cluster_test.go
marians: Good catch

marians: In fact, I didn't add it to all tests cases, since it's not a required argument.

marians:
@kopiczko Fine now?

commands/create_cluster.go
kopiczko: Yes. At least `error.Error()`.

Issue #87, opened 2017-06-14
Config file should have permission 0600, but has 0644


marians:
#### Observed:

The configuration file we generate gets the permission mask `-rw-r--r--` which in theory allows read access to everyone.

Seen on Mac OS.

#### Expected:

Permission mask `-rw-------`.

We are setting the permissions to `0600` explicitly here https://github.com/giantswarm/gsctl/blob/master/config/config.go#L180

marians:
Kind of weird, since we explicitly set the permission of the config file to `0600` (on creation):

https://github.com/giantswarm/gsctl/blob/master/config/config.go#L180

marians:
Here is a test program I ran locally both on Mac OS and Ubuntu and got the expected results in both cases.

```go
package main

import (
	"fmt"
	"io/ioutil"
	"os"
)

func main() {
	path := "./testfile"
	err := ioutil.WriteFile(path,
		[]byte("This is some content"),
		os.FileMode(0600))
	if err != nil {
		panic(err)
	}

	fileInfo, statErr := os.Stat(path)
	if statErr != nil {
		panic(err)
	}

	fmt.Printf("Permissions: %o (%s)\n",
		fileInfo.Mode().Perm(),
		fileInfo.Mode().Perm())
}
```

Execution and output (Mac OS):

```bash
$ go run main.go && ls -la testfile
Permissions: 600 (-rw-------)
-rw-------    1 marian  staff    20B Jun 16 09:01 testfile
```

marians:
Another test program that first creates a directory structure and places the test file inside, just as it happens in gsctl. Again, this works as aéxpected and it doesn't reproduce the problem.

```go
package main

import (
	"fmt"
	"io/ioutil"
	"os"
)

func main() {
	dirPath := "./dir/subdir"
	os.MkdirAll(dirPath, 0700)

	path := "./dir/subdir/testfile"
	err := ioutil.WriteFile(path,
		[]byte("This is some content"),
		0600)
	if err != nil {
		panic(err)
	}

	fileInfo, statErr := os.Stat(path)
	if statErr != nil {
		panic(err)
	}

	fmt.Printf("Permissions: %o (%s)\n",
		fileInfo.Mode().Perm(),
		fileInfo.Mode().Perm())
}
```

xh3b4sd:
I think your problem is this: https://play.golang.org/p/-0a3TPn7P-.

marians:
[...]

PR #165, opened 2017-12-02
Ensure config file mode of 0600

marians:
Fixes https://github.com/giantswarm/gsctl/issues/87

The problem was that the creation using `file, fileErr := os.Create(ConfigFilePath)` simply did not set any permission.

Since users will now have config files with permission 0644 from current gsctl versions, an additional permission udpdate after each write access to the config file should ensure the correct mode.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -24,6 +24,9 @@ const (
 	// ConfigFileName is the name of the configuration file, without ending
 	ConfigFileName = "config"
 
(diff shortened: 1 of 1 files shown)


marians:
Here: https://github.com/giantswarm/gsctl/pull/165/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR253

Issue #23, opened 2017-03-01
'gsctl logout' for expired session yields "unexpected end of JSON input"


marians:
Current output for a `gsctl logout` in case of an expired token:

```
$ gsctl logout
Info: The client doesn't handle the API's 401 response yet.
Seeing this error likely means: The passed token was no longer valid.
Error details:
2017/03/01 15:18:16 unexpected end of JSON input
exit status 1
```

The expected behaviour would be to simply remove the local copy of the token and print a success message that the user is logged out.

This requires the client to be able to understand the response returned by the API.

marians:
Needs verification

marians:
Current situation:

```nohighlight
$ gsctl logout
Error in API request to logout: unexpected end of JSON input
```

PR #166, opened 2017-12-02
Fixing logout in case of expired/unknown token

marians:
Fixes https://github.com/giantswarm/gsctl/issues/23

First, this fixes the mockServer for the test case. The API in fact returns no body in case of a 401 error.

Then the error handling was bad, which isn't really clean now, but works.

When logging out with a bad or expired token, now gsctl behaves as if the logout was successful.


commands/logout.go
```diff
@@ -65,11 +66,16 @@ func logoutOutput(cmd *cobra.Command, extraArgs []string) {
...
-		switch err.Error() {
-		case "":
+		headline := ""
+		subtext := ""
+
+		switch {
+		case IsNotAuthorizedError(err):
+			fmt.Printf("You have logged out from endpoint %s.\n", color.CyanString(logoutArgs.apiEndpoint))
+			os.Exit(0)
+		case err.Error() == "":
```
kopiczko: This is strange. But I see it was like that.

commands/logout.go
```diff
@@ -105,16 +111,18 @@ func logout(args logoutArguments) error {
...
 	if err != nil {
-		return fmt.Errorf("Error in API request to logout: %s", err.Error())
+		// special treatment for HTTP 401 (unauthorized) error,
+		// in which case no JSON body is returned.
+		if apiResponse.Response.StatusCode == http.StatusUnauthorized {
+			return microerror.Mask(notAuthorizedError)
+		}
+
+		// other cases
+		return microerror.Mask(errors.New("Error in API request to logout: " + err.Error()))
```
kopiczko: According to our guidelines you should create an error variable and mask it. Also we start error messages with lower letter.

commands/logout.go
```diff
@@ -105,16 +111,18 @@ func logout(args logoutArguments) error {
...
 	}
 
 	if logoutResponse.StatusCode != apischema.STATUS_CODE_RESOURCE_DELETED {
-		if apiResponse.Response.StatusCode == http.StatusUnauthorized {
-			// we ignore a 401 (Unauthorized) response here, as it means in most cases
-			// that the token submitted was already expired.
-			return nil
-		}
-		return fmt.Errorf("Error in API request to logout: %#v", logoutResponse)
+		return microerror.Mask(fmt.Errorf("Error in API request to logout: %#v", logoutResponse))
```
kopiczko: Ditto error

marians:
Changed

commands/logout.go
```diff
@@ -105,16 +110,18 @@ func logout(args logoutArguments) error {
...
 	}
 
 	if logoutResponse.StatusCode != apischema.STATUS_CODE_RESOURCE_DELETED {
-		if apiResponse.Response.StatusCode == http.StatusUnauthorized {
-			// we ignore a 401 (Unauthorized) response here, as it means in most cases
-			// that the token submitted was already expired.
-			return nil
-		}
-		return fmt.Errorf("Error in API request to logout: %#v", logoutResponse)
+		return microerror.Maskf(unspecifiedAPIError, fmt.Sprintf("Error in API request to logout: %v", logoutResponse))
```
kopiczko: Your error message will look like: `unspecified API error: Error in API request to logout: ...`. I'd rather make it something like`unspecified API error: requesting to logout: ...`. That would also solve the problem of capital letter. WDYT?

kopiczko: We talked about it. Just to reference: `return microerror.Maskf(unspecifiedAPIError, "requesting logout: %v", logoutResponse)`

Issue #24, opened 2017-03-03
'gsctl create keypair' should output expiry in days when convenient


marians:
Deriving the actual number of days from this output isn't easy in one's head:

```
New key-pair created with ID 722e93168… and expiry of 720 hours
```

It would be better to convert that into days (if hours > 24).

marians:
Also affects `gsctl create kubeconfig`

marians:
Expiry output is now friendlier and expressed in hours, days, weeks, months, ... Merged to `master` and will be released soon.

PR #167, opened 2017-12-04
Human-friendly duration output for key pair expiry

marians:
Fixes https://github.com/giantswarm/gsctl/issues/24, based on https://github.com/giantswarm/gsctl/pull/161

Also 
- aligns the output of `create kubeconfig` and `create keypair`.
- removes the output of line `Creating new key pair…`

### DIFF
--- a/commands/create_keypair.go
+++ b/commands/create_keypair.go
@@ -98,8 +98,9 @@ func addKeypair(cmd *cobra.Command, args []string) {
 	}
 
 	if apiResponse.StatusCode == 200 {
-		cleanID := util.CleanKeypairID(keypairResponse.Id)
-		msg := fmt.Sprintf("New key pair created with ID %s and expiry of %v hours", cleanID, keypairResponse.TtlHours)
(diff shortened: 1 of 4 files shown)


commands/create_keypair.go
```diff
@@ -98,8 +98,9 @@ func addKeypair(cmd *cobra.Command, args []string) {
 	}
 
 	if apiResponse.StatusCode == 200 {
-		cleanID := util.CleanKeypairID(keypairResponse.Id)
-		msg := fmt.Sprintf("New key pair created with ID %s and expiry of %v hours", cleanID, keypairResponse.TtlHours)
+		msg := fmt.Sprintf("New key pair created with ID %s and expiry of %v",
+			util.Truncate(util.CleanKeypairID(keypairResponse.Id), 10),
+			util.DurationPhrase(int(keypairResponse.TtlHours)))
```
kopiczko: You could use https://golang.org/pkg/time/#Duration.String

marians: That's a bit less friendly to read in my opinion

Thanks!

PR #170, opened 2017-12-04
Redact password argument value

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/2108

This PR should eliminate passwords from being transmitted via the command line header with API calls.


commands/commands.go
```diff
@@ -125,10 +125,30 @@ func randomRequestID() string {
...
+			args[index+1] = "REDACTED"
+		} else if len(args) > 1 && args[1] == "login" {
+			// this will explicitly only apply to the login command
+			if strings.HasPrefix(arg, "-p=") {
+				args[index] = "-p=REDACTED"
+			} else if arg == "-p" {
+				args[index+1] = "REDACTED"
+			}
+		}
+	}
```
xh3b4sd: There should actually be tests for such things.

commands/commands.go
```diff
@@ -125,10 +125,30 @@ func randomRequestID() string {
...
+
+	fmt.Printf("Args: %#v\n", args)
```
xh3b4sd: I think this should be removed before merging.

marians:
- Added a test
- Removed that print command
- Turned off CircleCI again, as it was never working anyway

marians:
@xh3b4sd Do you approve?

Issue #81, opened 2017-06-12
Files from expired key pairs should be removed


marians:
When `gsctl create kubeconfig` or `gsctl create keypair` is used repeatedly for the same cluster, certificate files start piling up in the certs folder. We should removed expired key pair files.

When fetching key pairs for a cluster, we get only those that are not yet expired. We could use that information to remove those keys and certificates that belong to unexpired key pairs.

Alternatively we could try and inspect the certificate like openssl does to read out the expiration date.

marians:
Solved in #171

PR #171, opened 2017-12-05
Key pair garbage collection

marians:
Solves #81


config/config.go
```diff
@@ -273,6 +278,13 @@ func Initialize(configDirPath string) error {
 
 	KubeConfigPaths = getKubeconfigPaths(HomeDirPath)
 
+	// apply garbage collection
+	randSource := rand.NewSource(time.Now().UnixNano())
+	randGenerator := rand.New(randSource)
+	if randGenerator.Float32() < garbageCollectionLikelihood {
+		GarbageCollectKeyPairs()
+	}
```
xh3b4sd: I think we should just do it always.

marians:
Just pulled back the review requests. Found that I still need to clean up a little.

config/garbage_collect.go
```diff
@@ -0,0 +1,85 @@
...
+	"os"
+	"strings"
+	"time"
+
+	"github.com/giantswarm/microerror"
+)
+
+// GarbageCollectKeyPairs removes files from expired key pairs
+func GarbageCollectKeyPairs() error {
+	files, err := ioutil.ReadDir(CertsDirPath)
```
xh3b4sd: I would use a walk function even if there is no nesting at the moment.

config/garbage_collect_test.go
```diff
@@ -0,0 +1,111 @@
...
+RvOTa8hYiU6A475WuZKyEHcwnGYe57u2I2KbMgcKjPniocj4QzgYsVAVKW3IwaOh
+yE+vPxsiUkvQHdO2fojCkY8jg70jxM+gu59tPDNbw3Uh/2Ij310FgTHsnGQMyA==
+-----END CERTIFICATE-----`
+
+	expired, err := isCertExpired([]byte(certPEM))
+	if err != nil {
+		t.Error(err)
+	}
+
+	if expired != true {
```
xh3b4sd: When the method should be verified the negative case is missing.

config/garbage_collect_test.go
```diff
@@ -0,0 +1,111 @@
...
+		t.Error("Expected true, got false")
+	}
+}
+
+func TestGarbageCollectKeyPairs(t *testing.T) {
```
xh3b4sd: We usually tend to structure test case names a bit like `Test_GarbageCollection_KeyPairs` for better readability.

config/garbage_collect_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// temporary config dir
+	dir, tempConfigErr := tempConfig("")
+	if tempConfigErr != nil {
+		t.Error(tempConfigErr)
+	}
+	defer os.RemoveAll(dir)
+
+	// copy test files over to temporary certs dir
+	basePath := "testdata"
+	files, _ := ioutil.ReadDir(basePath)
```
xh3b4sd: As I already mentioned we should aim to use abstractions for dependencies like filesystems.

marians:
To @xh3b4sd @kopiczko: CircleCI is activated but cannot be green currently, because https://github.com/giantswarm/gsctl/pull/109 isn't merged yet.

Thanks in advance for reviewing!

marians:
Added a test for isCertExpired for the opposite case. The other suggestions will have to wait for some indefinite time. Thanks!

Issue #129, opened 2017-10-19
add id or shortcut for endpoints


calvix:
New endpoint system is cool, but I find it weird that I need to use the whole URL when I want to change to endpoint.

current state:
```
 #: gsctl list endpoint
ENDPOINT URL                                          EMAIL                 SELECTED  LOGGED IN
https://api.g8s.gauss.eu-central-1.aws.gigantic.io    vaclav@giantswarm.io  no        yes
https://api.g8s.ginger.eu-central-1.aws.gigantic.io   vaclav@giantswarm.io  no        yes
https://api-g8s.giantswarm.io                         vaclav@giantswarm.io  no        yes
[calvix@nitro] ~ | 
 #: gsctl select endpoint https://api-g8s.giantswarm.io 
Endpoint selected: https://api-g8s.giantswarm.io
```

I would like something like:
```
 #: gsctl list endpoint
ENDPOINT ID             ENDPOINT URL                                          EMAIL                 SELECTED  LOGGED IN
1                       https://api.g8s.gauss.eu-central-1.aws.gigantic.io    vaclav@giantswarm.io  no        yes
2                       https://api.g8s.ginger.eu-central-1.aws.gigantic.io   vaclav@giantswarm.io  no        yes
3                       https://api-g8s.giantswarm.io                         vaclav@giantswarm.io  no        yes
[calvix@nitro] ~ | 
 #: gsctl select endpoint 3
Endpoint selected: https://api-g8s.giantswarm.io
```

or  with shortcut
```
ENDPOINT ID             ENDPOINT URL                                          EMAIL                 SELECTED  LOGGED IN
gauss                   https://api.g8s.gauss.eu-central-1.aws.gigantic.io    vaclav@giantswarm.io  no        yes
ginger                  https://api.g8s.ginger.eu-central-1.aws.gigantic.io   vaclav@giantswarm.io  no        yes
lycan                   https://api-g8s.giantswarm.io                         vaclav@giantswarm.io  no        yes
[calvix@nitro] ~ | 
 #: gsctl select endpoint lycan
Endpoint selected: https://api-g8s.giantswarm.io
```

Shortcut could be defined when adding endpoint via flag
```
gsctl -e  https://api-g8s.giantswarm.io -n lycan login vaclav@giantswarm.io
```
if not specified we could either provide some random code word from predefined list, assign number or make it required flag

calvix:
WDYT guys @marians @oponder

marians:
[...]

PR #180, opened 2017-12-12
Allow aliases for endpoints

marians:
This is a first step towards #129 

Users can manually add an `alias` key to an endpoint entry in the `config.yaml` file. Example:

```yaml
endpoints:
  http://localhost:9000:
    email: email@example.com
    token: ...
    alias: local
...
```

As soon as there is at least one endpoint with alias in the config, the `gsctl list endpoints` command shows an additional `ALIAS` column and is sorted by alias:

```nohighlight
ALIAS  ENDPOINT URL                                          EMAIL                    SELECTED  LOGGED IN
local  http://localhost:9000                                 developer@example.com    no        yes
n/a    https://api.g8s.foobar.eu-central-1.aws.gigantic.io   developer@example.com    yes       yes
```

With this alias, users can `gsctl select endpoint local`.

In addition, a line is added to the output of `gsctl info` when an alias is set for the current endpoint.

Note: As soon as you run an older `gsctl` version, any `alias` keys will be removed from your config file.

Note: So far, uniqueness of aliases is not verified. This would be something to take care of when suporting the adding/manipulation of an alias.


commands/info_test.go
```diff
@@ -9,12 +9,52 @@ import (
 // Test_PrintInfo simply executes the printInfo function.
 // TODO: actually test what this does
 func Test_PrintInfo(t *testing.T) {
+	// our test config YAML
+	yamlText := `last_version_check: 0001-01-01T00:00:00Z
+updated: 2017-09-29T11:23:15+02:00
+endpoints:
+  https://myapi.domain.tld:
```
xh3b4sd: Not related to this PR but just to have that said. IMO there should either be an explicit ID to identify an endpoint, or none at all. If there is an ID it should be the codename IMO. If there is none the data structure should be a list of objects. IMO the API endpoint of an installation is just one property of many an installation can have. IMO it is not correct to use the API endpoint as identifier.

commands/list_endpoints.go
```diff
@@ -61,30 +61,59 @@ func endpointsTable(args listEndpointsArguments) string {
...
-	}
-
 	// get keys (URLs) and sort by them
 	endpointURLs := make([]string, 0, len(config.Config.Endpoints))
 	for u := range config.Config.Endpoints {
 		endpointURLs = append(endpointURLs, u)
 	}
 
+	// detect if we want to show the alias column
+	haveAlias := false
```
xh3b4sd: Grammar nitpick, this should be `hasAlias` IMO.

commands/list_endpoints.go
```diff
@@ -61,30 +61,59 @@ func endpointsTable(args listEndpointsArguments) string {
...
+	for _, endpoint := range endpointURLs {
+		if config.Config.Endpoints[endpoint].Alias != "" {
+			haveAlias = true
+		}
+	}
```
xh3b4sd: A good practise is to have a dedicated function for such things. It is reusable and easer to use in the code.

commands/list_endpoints.go
```diff
@@ -61,30 +61,59 @@ func endpointsTable(args listEndpointsArguments) string {
...
 	// get keys (URLs) and sort by them
 	endpointURLs := make([]string, 0, len(config.Config.Endpoints))
 	for u := range config.Config.Endpoints {
 		endpointURLs = append(endpointURLs, u)
 	}
 
+	// detect if we want to show the alias column
+	haveAlias := false
+	for _, endpoint := range endpointURLs {
+		if config.Config.Endpoints[endpoint].Alias != "" {
```
xh3b4sd: Accessing the map item with `config.Config.Endpoints[endpoint]` is done quite often. This could be improved. It is also that this code can panic in case the map does not hold what you need. Even if you validate this above, this quite big function relies on the validation in a lot of places and this can go wrong quite quickly based on my personal experiences.

config/config.go
```diff
@@ -138,13 +141,33 @@ func (c *configStruct) StoreEndpointAuth(endpointURL string, email string, token
...
+	ep := ""
+
+	// first check if the endpointURL matches an alias.
+	argumentIsAlias := false
+	for key := range c.Endpoints {
+		if endpointAliasOrURL == c.Endpoints[key].Alias {
+			argumentIsAlias = true
+			ep = key
+		}
+	}
```
xh3b4sd: A method like `Endpoints.GetByKey(key string) (string, error)` would make sense to me here. In case there is none, a `notFoundError` should be returned and the according matcher could be used below.

marians:
Thanks for the review! I'll leave the room for improvement mostly open.

PR #189, opened 2018-01-19
Improve `show cluster` error handling

tuommaki:
Handle error cases where API returns non-OK HTTP error code:
  - HTTP 401 Unauthorized
    - Probably doesn't occure often in the real life as login has failed
      beforehand, but in case it happens, handle this gracefully with
      meaningful error message.
  - HTTP 404 Not Found
    - If non-existing cluster id is given for `show cluster`, handle API
      response gracefully and give meaningful error message.
  - HTTP 500 Internal Server Error
    - When all hope is lost, try to still behave sensibly for the end
      user. It's the API that has something wrong, not `gsctl`.

While here, also added check to showClusterRunOutput that checks length
of the Workers slice before addressing the first element. Earlier this
gave "index out of range" error especially when API errors were not
handled. Now it behaves gracefully.

### DIFF
--- a/commands/show_cluster.go
+++ b/commands/show_cluster.go
@@ -118,10 +118,20 @@ func getClusterDetails(clusterID, token, endpoint string) (gsclientgen.V4Cluster
 		return result, microerror.Mask(couldNotCreateClientError)
 	}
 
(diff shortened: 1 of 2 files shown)


marians:
Very much appreciated!

Issue #187, opened 2018-01-09
Command to upgrade a cluster


marians:
Part of https://github.com/giantswarm/giantswarm/issues/2022

To manually trigger an upgrade of a guest cluster to a newer version, gsctl needs a command.
 
The backend automatically selects to which version the cluster will be upgraded. The logic for that can be replicated like this:

- Pick the lowest version number that is higher than the current version number.
- Pick only from active releases.

This can be used to display to the user to which version we'll upgrade and what the changes are.

In addition, a confirmation should be required by default, overridable using `-f/--force`.

marians:
The most straightforward way to name the command:

    gsctl upgrade cluster

Details

- Details of the target version and a basic comparison between the current and the target version as well as the target version's changelog will be displayed. See mockup below.
- A confirmation will be required unless the `-f` or `--force` flag has been added.

(WIP, will be edited)

oponder:
Looks good to me! I am missing the "warning" about what the consequences of the upgrade are (workers will be drained, etc). Are we consciously excluding that step from `gsctl`?

marians:
Yes, good that you ask. With the general assumption of gsctl being the powertool for knowledgeable users (who feel that happa might be too chatty, to verbose, too slow), I decided to not show that. It would make the confirmation step content really big. And I really don't want a two step confirmation in gsctl.

oponder:
IMHO, powertools should still have safety features or some warnings on them.

I think the user must at least be aware of these two things:

1. Make sure your services still run with one node fully drained.
2. The kubernetes API will be unavailable for some time during the upgrade.

Perhaps we can add just these two notices?

Alternatively, I came up with an idea: what if we temporarily add a worker as part of the upgrade process.

That way we create the necessary space to successfully perform an upgrade. I know this is very not in scope of this discussion though.

oponder:
Reminded me of this woodworking video I saw. The kid asks "Why does the start button have a cover over it, but the stop button doesn't"?

Then the dad says "So you can't accidentally hit the start button, but you want to be able to hit the stop button very quickly"

<img width="903" alt="screen shot 2018-01-10 at 17 20 20" src="https://user-images.githubusercontent.com/455309/34783086-daa0485a-f62a-11e7-82d0-0a16b8d94e4c.png">

[...]

PR #192, opened 2018-01-24
New 'upgrade cluster' command

marians:
Solves https://github.com/giantswarm/gsctl/issues/187

### Depends on

- https://github.com/giantswarm/gsclientgen/pull/23
- https://github.com/giantswarm/giantswarm/issues/2343

### TODO

- [x] Fix URL for upgrade docs (if needed)
- [x] Adapt to changes in https://github.com/giantswarm/gsctl/pull/197


commands/upgrade.go
```diff
@@ -0,0 +1,16 @@
...
+import "github.com/spf13/cobra"
+
+var (
+	// UpgradeCommand is the command to upgrade things
+	UpgradeCommand = &cobra.Command{
+		Use:   "upgrade",
+		Short: "Upgrade clusters",
+		Long:  `Lets you upgrade a cluster`,
+	}
+)
```
kopiczko: Those var block look weird in this situation.

```
// UpgradeCommand is the command to upgrade things.
var UpgradeCommand = &cobra.Command{
	Use:   "upgrade",
	Short: "Upgrade clusters",
	Long:  `Lets you upgrade a cluster`,
}
```

It is more concise, less indentation.

But not a blocker.

Also full stop is missing.

commands/upgrade_cluster.go
```diff
@@ -0,0 +1,344 @@
...
+			fmt.Printf("    - %s: %s\n", change.Component, change.Description)
+		}
+
+		fmt.Println("")
+		fmt.Println("NOTE: Upgrading may impact your running workloads and will make the cluster's")
+		fmt.Println("Kubernetes API unavailable temporarily. Before upgrading, please acknowledge the")
+		fmt.Println("details described in")
+		fmt.Println("")
+		fmt.Printf("    %s\n", upgradeDocsURL)
+		fmt.Println("")
```
kopiczko: Why not printing the changelog and note when args.force?

kopiczko (approved):

commands/upgrade_cluster.go
marians: The idea was that if the user uses -f/--force, that information couldn't really be used anyway.

Now that I'm thinking about it, it might still be that the user wants to scroll back and find out what really happened. And -f/--force is useful for automations, where that output would still be stored somewhere. OK, I'll change that.

commands/upgrade.go
marians: If there was only one import, should I then also write import "foo" instead of a block? I thought Gophers prefer diffs to be minimal, hence all the fuzz about trailing commas in blocks like

```go
myslice := []string{
    s1,
    s22,
}
```

kopiczko: > If there was only one import, should I then also write import "foo" instead of a block?

I don't think that improves readability, it also doesn't reduce indentation.

What I mean

```
var (
	someString = "abc"
	someInt    = 1
)
```

^ That improves readability. Also groups related vars together.

```
var (
	someStruct = type {
		a: 1,
	}
)
```

^ That doesn't improve readability.

Trailing commas are different story. They improve diffs.

kopiczko: I don't think it's so important to spend more time on that. Do what you think is right.

commands/upgrade_cluster.go
kopiczko: Yeah, when you don't care about the output `> /dev/null`.

commands/upgrade.go
marians: I'll leave it consistent with the rest of the command files then.

commands/upgrade_cluster.go
marians: Changed

commands/upgrade_cluster.go
```diff
@@ -0,0 +1,344 @@
...
+	if releasesErr != nil {
+		return result, microerror.Mask(releasesErr)
+	}
+	releaseVersions := []string{}
+	for _, r := range releasesResult.releases {
+		releaseVersions = append(releaseVersions, r.Version)
+	}
+
+	// define the target version to upgrade to
+	targetVersion := successorReleaseVersion(details.ReleaseVersion, releaseVersions)
```
oponder: Some of our older clusters have no release version because they were made before we had that feature.

Those clusters should also be considered not upgradeable I think. 

Not totally sure what this code will do in case `details.ReleaseVersion` == `""`

TheoBrigitte:
Any update on this?

It would be quite handy to have `gsctl upgrade`.

marians:
Yeah, I'll make it available for admins only

marians:
Actual code reviews have been a long time ago. Since then, the client implementation has changed drastically. Happy for a proper review.

commands/error.go
```diff
@@ -437,6 +439,28 @@ func IsUnspecifiedAPIError(err error) bool {
 	return microerror.Cause(err) == unspecifiedAPIError
 }
 
+// noUpgradeAvailableError means that the user wanted to start an upgrade, but
+// there is no newer version available for the given cluster
+var noUpgradeAvailableError = &microerror.Error{
+	Kind: "noUpgradeAvailableError",
+	Desc: "no upgrade available for the current version",
```
xh3b4sd: IMO these are supposed to be sentences. See https://godoc.org/github.com/giantswarm/microerror#Error.

commands/upgrade_cluster.go
```diff
@@ -0,0 +1,311 @@
...
+
+	"github.com/giantswarm/gsctl/config"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	// upgradeClusterActivityName assigns API requests to named activities
+	upgradeClusterActivityName = "upgrade-cluster"
+
+	upgradeDocsURL = "https://docs.giantswarm.io/reference/cluster-upgrades/"
```
xh3b4sd: I remember you saying you wanted to drop the docs link for now. Don't remember the reasons.

commands/upgrade_cluster.go
```diff
@@ -0,0 +1,311 @@
...
+	// return first item that is greater than version
+	comp := semver.New(version)
+	for _, v := range versions {
+		vv := semver.New(v)
+		if comp.LessThan(*vv) {
+			return v
+		}
+	}
+
+	return ""
```
xh3b4sd: IMO here should an error be returned. Implying behaviour by a certain content difference is an anti pattern IMO.

xh3b4sd (commented):
I am not sure what a proper review here means since this project is so dramatically different to most of the others. Just left some comments where I felt I could say something anyway.

commands/upgrade_cluster.go
kopiczko: For me is a blocker also. Two tings: what if this is newest version? What if old cluster doesn't have a version (credits to Oli)? There are maybe even more corner cases. I prefer having error than random stuff happening.

xh3b4sd:
What can we do to get this thing here merged? It would be pretty useful I guess and it is open for a pretty long time now.

marians:
Looks like I have to make the tests work again

commands/upgrade_cluster.go
marians: Sorry to have ignored this old comment. Not sure why... I have added treatment for this case to `successorReleaseVersion()`.

PR #193, opened 2018-01-24
Changes certificate-organizations flag to groups

JosephSalisbury:
I can never remember `--certificate-organizations`, so here's a shorthand!

e.g:
```
Flags:
  -o, --certificate-organizations string   A comma separated list of organizations for the issued certificates 'O' fields.
  -c, --cluster string                     ID of the cluster
```

### DIFF
--- a/commands/create_kubeconfig.go
+++ b/commands/create_kubeconfig.go
@@ -180,7 +180,7 @@ func init() {
 	CreateKubeconfigCommand.Flags().StringVarP(&cmdCNPrefix, "cn-prefix", "", "", "The common name prefix for the issued certificates 'CN' field.")
 	CreateKubeconfigCommand.Flags().StringVarP(&cmdKubeconfigSelfContained, "self-contained", "", "", "Create a self-contained kubectl config with embedded credentials and write it to this path.")
 	CreateKubeconfigCommand.Flags().StringVarP(&cmdKubeconfigContextName, "context", "", "", "Set a custom context name. Defaults to 'giantswarm-<cluster-id>'.")
-	CreateKubeconfigCommand.Flags().StringVarP(&cmdCertificateOrganizations, "certificate-organizations", "", "", "A comma separated list of organizations for the issued certificates 'O' fields.")
+	CreateKubeconfigCommand.Flags().StringVarP(&cmdCertificateOrganizations, "certificate-organizations", "o", "", "A comma separated list of organizations for the issued certificates 'O' fields.")
 	CreateKubeconfigCommand.Flags().BoolVarP(&cmdForce, "force", "", false, "If set, --self-contained will overwrite existing files without interactive confirmation.")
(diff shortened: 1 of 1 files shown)


marians:
@JosephSalisbury Merging this as is would definitely be a UX improvement.

However...

As for some consistency questions, I am wondering whether we should be concerned that `-o` means something different in the context of another command, `gsctl create cluster`, where it is the shorthand for `--owner`.

As an alternative approach, we could now decide to rename `--certificate-organizations`, because

- it's in fact really hard to remember and 
- also not quite accurate, plus it has the term "organizations" in it, which is a concept we use differently in Giant Swarm.

In Kubernetes' RoleBinding resources, the resource type `group` is used. So it might make sense to use that term here, too.

Suggestion:

- long version `--cert-groups`
- if we still need a short version, let's make it `-g`.

Sorry for making it complicated here. That's my job.

marians:
Ping @giantswarm/sig-ux Any opinions?

oponder:
How about just `--groups` and `-g`?

The `cert` part isn't needed since we're creating a cert. I know I was the one who came up with `certificate-organizations` btw, and I regret it :P

JosephSalisbury:
I was under the impression that `system:masters` was part of the O field in the certificate, which is why I went with `-o`.

I am not particularly fussed what the shorthand is - if you guys are happy with `-g` and `--groups`, I can update here.

marians:
@JosephSalisbury You are right, in the x509 terminology, the value ends up here as part of the certificate:

```nohighlight
Subject: O=system:masters, CN=email@example.com.user.api.vxvc7.k8s.codename.eu-central-1.aws.gigantic.io
           ^^^^^^^^^^^^^^
```

Still thinking...

oponder:
Thats the road I went down at first. But the thing is the O field gets interpreted as the `group` that the user belongs to, and that is what we care about truly.

JosephSalisbury:
That is a good enough argument for me. Changing here to `--groups`/`-g`.

JosephSalisbury:
@marians @oponder @puja108 RFR

oponder:
This might need a lil customer announcement when it is released.

I think I'll do a similar terminology on Happa.

Also, instead of replacing the flag, we should deprecate the old and introduce the new.

We should also do the same thing at the API level:

https://docs.giantswarm.io/api/#operation/addKeyPair

marians:
@JosephSalisbury Note that this is now a breaking change. I'd appreciate the introduction of a hidden flag `--certificate-organizations` for downwards compatibility that just gets copied over to the `cmdCertificateOrganizations` variable. You can use func `defaultCreateKubeconfigArguments` for that.

And: the `gsctl create keypair` command has the `--certificate-organizations` flag, too, and requires the same changes. Yes, there is quite some duplication going on.

marians:
@JosephSalisbury What do you want to do with this PR? Should we close it?

JosephSalisbury:
Yeah, I just put this into my history a while ago and gave up.

Issue #196, opened 2018-02-15
Connection problems should be handled gracefully


marians:
Now that some gsctl users are behind stricter firewalls, network level blocking may apply. It would be great if commands like `gsctl login` would handle these in a specific and helpful way.

### What did you do/observe?

```
$ gsctl login dev@giantswarm.io
Password for dev@giantswarm.io on https://gs-api.g8s.eu-central-1.aws.domain.com:
invalid character '<' looking for beginning of value
```

### What did you expect?

Something like

```
Cannot connect to <api-endpoint>.
Please make sure you have a working connection and try `gsctl ping` to verify.
```

marians:
The error `invalid character '<' looking for beginning of value` we are handling (or rather printing) there originates from unsuccessful JSON unmarshaling here in the client:

https://github.com/giantswarm/gsclientgen/blob/master/default_api.go#L833

The client doesn't seem to handle connection problems in a specific way. We might have to fix that upstream in https://github.com/swagger-api/swagger-codegen

marians:
Actually its rather simple. We just have to check the HTTP response status code.

marians:
That's done

PR #197, opened 2018-02-15
Improve and align error handling, including 404 Forbidden error, in many commands

marians:
For https://github.com/giantswarm/gsctl/issues/196

Preview:

```
$ gsctl login dev@example.io
Password for dev@example.io on <endpoint>:
Access Forbidden
The client has been denied access to the API endpoint with an HTTP status of 403.
Please make sure that you are in the right network or VPN. Once that is verified,
check back with Giant Swarm support that your network is permitted access.
```

### Commands touched:

- `create cluster`
  - move out handling of common errors
  - handle 403 forbidden error
- `create kubeconfig`
  - move out handling of common errors
  - handle 403 forbidden error
- `create keypair`
  - refactored to match command blueprint, applied structure according to `create kubeconfig`
  - applied same client timeout as for `create kubeconfig`
  - handle 403 Forbidden error
- `delete cluster`
  - move out handling of common errors
  - handle 403 forbidden error
- `list keypairs`
  - move out handling of common errors
  - handle 403 forbidden error
- `login`:
  - move out handling of common errors
  - handle 403 forbidden error
- `logout`
  - handle 403 forbidden error
- `ping`
  - move out handling of common errors
  - handle 403 forbidden error
- `scale cluster`
  - move out handling of common errors
  - handle 403 forbidden error

### DIFF
--- a/commands/create_keypair.go
+++ b/commands/create_keypair.go
@@ -196,6 +196,9 @@ func createKeypair(args createKeypairArguments) (createKeypairResult, error) {
 		addKeyPairActivityName, cmdLine)
 
 	if err != nil {
+		if apiResponse.Response != nil && apiResponse.Response.StatusCode == http.StatusForbidden {
+			return result, microerror.Mask(accessForbiddenError)
+		}
 		return result, microerror.Mask(err)
 	}
 
--- a/commands/create_keypair.go
+++ b/commands/create_keypair.go
@@ -1,9 +1,10 @@
 package commands
 
 import (
-	"errors"
 	"fmt"
+	"net/http"
 	"os"
+	"time"
 
 	"github.com/fatih/color"
 	"github.com/giantswarm/gsclientgen"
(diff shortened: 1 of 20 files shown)


marians:
@kopiczko Can I kindly confront you with this?

commands/create_keypair.go
```diff
@@ -42,79 +89,138 @@ func init() {
...
-func checkAddKeypair(cmd *cobra.Command, args []string) error {
+func createKeyPairPreRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
+	args := defaultCreateKeypairArguments()
+	err := verifyCreateKeypairPreconditions(args)
 
-	endpoint := config.Config.ChooseEndpoint(cmdAPIEndpoint)
-	token := config.Config.ChooseToken(endpoint, cmdToken)
+	if err == nil {
+		return
+	}
```
kopiczko: In other function we have

```
+	if err != nil {
+		handleCommonErrors(err)
```

Even though I'm **huge** fan of fast returns, in this case, this is possible there will be something added forcing use of additional indentation and obscuring the diff. I think unifying that would be beneficial.

docs/Command-Blueprint.md
```diff
@@ -122,15 +134,17 @@ func verbNounRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
 	result, err := verbNoun()
```
kopiczko: How does that work w/o an argument?

docs/Command-Blueprint.md
```diff
@@ -149,6 +163,49 @@ func verbNounRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
...
+		}
+
+		return result, microerror.Mask(err)
+	}
+
+  // populate result base on some response information etc.
+  result.someAttribute = someResponse.someValue
 
+  return result, nil
 }
```
kopiczko: `go fmt ./...`

marians: As this is a markdown file, that doesn't work

docs/Command-Blueprint.md
marians: This is only markdown (example code)

docs/Command-Blueprint.md
kopiczko:
```
// verbNoun performs our actual function. It usually creates an API client,
// configures it, configures an API request and performs it.
func verbNoun(args verbNoundArguments) (verbNounResult, error) {
	result := verbNounResult{}

	// prepare client
	clientConfig := client.Configuration{
		Endpoint:  args.apiEndpoint,
		Timeout:   10 * time.Second,
		UserAgent: config.UserAgent(),
	}
	apiClient, clientErr := client.NewClient(clientConfig)
	if clientErr != nil {
		return result, microerror.Mask(couldNotCreateClientError)
	}

	authHeader := "giantswarm " + args.token
	someResponse, rawResponse, err := apiClient.DoSomething(authHeader,
		requestIDHeader, verbNounActivityName, cmdLine)

	if rawResponse == nil || rawResponse.Response == nil {
		return result, microerror.Mask(noResponseError)
	}

	// handle request errors
	if err != nil {

		switch rawResponse.StatusCode {
		case http.StatusNotFound:
			return result, microerror.Mask(clusterNotFoundError)
		case http.StatusUnauthorized:
			return result, microerror.Mask(notAuthorizedError)
		case http.StatusForbidden:
			return result, microerror.Mask(accessForbiddenError)
		}

		if rawResponse.StatusCode >= 500 {
			return result, microerror.Maskf(internalServerError, err.Error())
		}

		return result, microerror.Mask(err)
	}

	// populate result base on some response information etc.
	result.someAttribute = someResponse.someValue

	return result, nil
}
```

docs/Command-Blueprint.md
kopiczko: I didn't notice :D

commands/create_keypair.go
kopiczko: I still think this comment is valid.

marians: Still thinking about doing the alignment...

marians: I would actually rather align to fast returns in all commands, as it would decrease the complexity of the code involved.

marians:
@kopiczko To commit https://github.com/giantswarm/gsctl/pull/197/commits/d011cd06355070767d30459c82f88943648d44be and acting against your recommendation:

The function `*PreRunOutput` has the purpose to print errors and exit only if errors occurred. So I think the `if err == nil { return }` logic is a perfect fit. I don't see this change in the near future.

kopiczko:
If `if err == nil { return }` works everywhere that's even better.

PR #204, opened 2018-03-06
Show KVM port mappings if they are present.

oponder:


oponder:
This is what it looks like:

```
$ gsctl show cluster 12345
ID:                            12345
Name:                          My Cluster
Created:                       2017 Nov 08, 14:38 UTC
Organization:                  giantswarm
Kubernetes API endpoint:       https://...
Release version:               n/a
Workers:                       ..
CPU cores in workers:          ..
RAM in worker nodes (GB):      ..
Storage in worker nodes (GB):  ..
Ingress Ports:                 http:  30071
                               https: 30072
```

Gopkg.lock
```diff
@@ -38,10 +38,10 @@
   version = "v2.0.2"
 
 [[projects]]
-  branch = "master"
+  branch = "add-kvm-port-mappings"
```
oponder: TODO: change back to master once https://github.com/giantswarm/gsclientgen/pull/25 is merged.

marians:
As for the output format:

```
Ingress Ports:                 http:  30071
                               https: 30072
```

I would prefer if we could stick to the non-hierarchical format we currently have. Like this:

```
Storage in worker nodes (GB):  ..
Ingress port for http:         30071
Ingress port for https:        30072
```

Reason: The output should be usable for simple shell scripting and processing with cut, awk and the likes.

Also please use lowercase p in "port".

oponder:
Ok! adjusted!

Issue #203, opened 2018-03-06
login command should handle inactive user account gracefully


marians:
### What did you do/observe?

```
$ gsctl login user@giantswarm.io -e https://api.g8s.codename.eu-central-1.aws.gigantic.io
Password for user@giantswarm.io on https://api.g8s.codename.eu-central-1.aws.gigantic.io:
Unhandled response code: 10017
```

### What did you expect?

Some useful error message

### Details

Code `10017` stands for an inactive user account. The likely cause is that the expiry date is reached.

marians:
Reference for the code meaning: https://github.com/giantswarm/api-schema/blob/master/api_schema.go#L28

PR #213, opened 2018-03-07
Handle inactive account

marians:
Fixes https://github.com/giantswarm/gsctl/issues/203


commands/login_test.go
```diff
@@ -164,3 +167,39 @@ selected_endpoint: "` + mockServer.URL + `"
...
+		email:       "developer@giantswarm.io",
+		password:    "test password",
+	}
+
+	_, err = login(args)
+	if err == nil {
+		t.Error("Expected error, got nil")
+	}
+	if !IsUserAccountInactiveError(err) {
+		t.Errorf("Expected userAccountInactiveError, got '%s'", err.Error())
```
TheoBrigitte: You could use `%q` here instead of `'%s'`, it will quote the string for you

marians:
Thanks @TheoBrigitte !

Issue #198, opened 2018-02-21
Help command shown after create a cluster is not useful


pipo02mix:
### What did you do/observe?
When you create a new cluster the output help text indicates 
```
gsctl create kubeconfig --cluster=<cluster-id>
```
But nowadays all clusters are with RBAC enabled, so the command creates a user that has not right to perform any action

### What did you expect?

Expose the command with the organization group and the ttl
```
gsctl create kubeconfig --cluster=<cluster-name> --certificate-organizations system:masters --ttl 1
```

marians:
As the second example creates a powerful key pair, we might want to explain a bit more.

- The key pair will have admin rights.
- Check the reference/help for more details on how to create a key pair with different privileges.

pipo02mix:
done

PR #219, opened 2018-04-02
Add an extra comment when cluster is created warning about the orgs flag

pipo02mix:
Towards https://github.com/giantswarm/gsctl/issues/198

### DIFF
--- a/commands/create_cluster.go
+++ b/commands/create_cluster.go
@@ -276,7 +276,7 @@ func createClusterExecutionOutput(cmd *cobra.Command, args []string) {
 		fmt.Printf("    %s\n\n", color.YellowString(fmt.Sprintf("gsctl create kubeconfig --cluster=%s \n\n", result.id)))
 		fmt.Println("Take into consideration all guest clusters have enabled RBAC and may you want to provide a correct organization for the certificates (like operators, testers, developer, ...)")
 		fmt.Println("")
-		fmt.Printf("    %s\n\n", color.YellowString(fmt.Sprintf("gsctl create kubeconfig --cluster=%s --certificate-organizations <group>", result.id)))
+		fmt.Printf("    %s\n\n", color.YellowString(fmt.Sprintf("gsctl create kubeconfig --cluster=%s --certificate-organizations system:masters", result.id)))
 		fmt.Println("")
 		fmt.Println("To know more about how to create the kubeconfig run")
 		fmt.Println("")
--- a/commands/create_cluster.go
+++ b/commands/create_cluster.go
@@ -273,7 +273,14 @@ func createClusterExecutionOutput(cmd *cobra.Command, args []string) {
 		}
 		fmt.Println("Add key pair and settings to kubectl using")
 		fmt.Println("")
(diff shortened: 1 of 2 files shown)


marians:
So this is a preview of what the new output would look like:

```
...
Add key pair and settings to kubectl using

    gsctl create kubeconfig --cluster=abc12

Take into consideration all guest clusters have enabled RBAC and may you want to provide a correct organization for the certificates

    gsctl create kubeconfig --cluster=abc12 --certificate-organizations client:operators
```

The `client:operators` string is just an example, right? I am pretty sure that some people will simply try to use that exact string without thinking about it and some others will wonder "why client:operators?".

I suggest as an alternative:

1. Use `system:masters` instead and explain what it means
2. Encourage users to use `gsctl create kubeconfig --help` to find out more about the options.

WDYT @pipo02mix @oponder ?

pipo02mix:
> 1. Use system:masters instead and explain what it means

I wanted to highlight you can use your own org. Indeed, I wanted to avoid people copy-pasting without thinking the consequences (admin access) but possibly the example used is not the best. What about `<team>:<testers>`

About second one I add an extra comment :)

oponder:
At first glance at it's current state, it confuses me a bit more than it helps I think.

Original problem is that the output does not help the user get further with accessing the cluster right? This new text won't really achieve that. I think we could show the sample command with `system:masters`, but educate about why you want to set up groups and roles with more fine grained permissions as soon as possible.

pipo02mix:
can you take a look again?

marians:
Checked, and it's looking good, except for the empty lines in the output.

```nohighlight
Requesting new cluster for organization 'otherorg'
New cluster with ID 'ioo86' for organization 'otherorg' is launching.
Add key pair and settings to kubectl using

    gsctl create kubeconfig --cluster=ioo86



Take into consideration all guest clusters have enabled RBAC and may you want to provide a correct organization for the certificates (like operators, testers, developer, ...)

    gsctl create kubeconfig --cluster=ioo86 --certificate-organizations system:masters


To know more about how to create the kubeconfig run

    gsctl create kubeconfig --help

```

In one case it's 3 empty lines, in another case it's 2. Let's make that 1 empty line at most.

marians:
Now there is a newline missing at the very end

marians:
... and after every command example.

```
Requesting new cluster for organization 'otherorg'
New cluster with ID '0r93p' for organization 'otherorg' is launching.
Add key pair and settings to kubectl using

    gsctl create kubeconfig --cluster=0r93p
Take into consideration all guest clusters have enabled RBAC and may you want to provide a correct organization for the certificates (like operators, testers, developer, ...)

    gsctl create kubeconfig --cluster=0r93p --certificate-organizations system:masters
To know more about how to create the kubeconfig run

    gsctl create kubeconfig --help
```

pipo02mix:
wait I am testing it, you are too quick! I made a commit in the Github UI but I realized I needed to test it. Checking in my local

pipo02mix:
ok now it is tested. PTAL

PR #227, opened 2018-04-19
Handle org not found error

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/2083

Preview:

```nohighlight
gsctl create cluster -n Test -o nonexisting
Requesting new cluster for organization 'nonexisting'
Organization not found
The organization set as owner does not exist, or you are not a member of it.
```


puja108 (approved):

marians:
Thanks for the in-depth review! No, I haven't tested that. Will have to launch my local setup agin...

marians:
So I checked. If the org exists, but the user is not a member of it, this happens:

```
gsctl create cluster -o giantswarm -n Test
Requesting new cluster for organization 'giantswarm'
You are not authorized for this action.
Please check whether you are logged in with the right credentials using 'gsctl info'.
```

Issue #137, opened 2017-11-09
Support shorter TTLs


oponder:
Our API takes the TTL for a keypair in hours: https://docs.giantswarm.io/api/#operation/addKeyPair

However in `gsctl create kubeconfig`, the `ttl` flag is always in days.

It would be nice to also support a smaller time interval. Usually I only need 1 hour in a customers cluster to verify an issue. 

Note this is not a bug, just a feature request for some way to create keypairs with a ttl of less than 1 day.

marians:
What do you think about allowing the use of the typical time unit syntax like `1h`, `1d`, `1w` etc?

We should make "hour" the default unit though, if no unit is given. This is of course a behaviour change that needs proper announcement.

oponder:
Sounds good to me! 

I'm thinking it might be better to error with a clear message if someone doesn't provide a unit. That way there is no ambiguity and people will learn soon enough to include the unit.

`"TTL is invalid. Please provide a unit. e.g. 1h, 1d, 1w"`

But no strong feelings about that

marians:
Throwing an error is also fine with me. It's more consistent and predictable that way.

marians:
This can be tested/reviewed in https://github.com/giantswarm/gsctl/pull/228

marians:
Done

PR #228, opened 2018-04-20
Allow key pair lifetime shorter than 1 day (breaking change)

marians:
Solves https://github.com/giantswarm/gsctl/issues/137

This is a breaking change for the usage of the `gsctl create <kubeconfig|keypair>` commands when applying the `--ttl` flag. Before, a number like `10` was accepted as a value. Now, the value must be a combination of a number and a unit, like `10d`.

Supported units are `h` for hour, `w` for week, `m` for month, and `y` for year.

The change is easy to adapt to. Simply add a `d` to the value as it was used before.

### Preview:

```
$ gsctl create kubeconfig -c 0r93p --ttl 1h
New key pair created with ID 7f3f88daf… and expiry of 1 hour
Certificate and key files written to:
/Users/marian/.config/gsctl/certs/0r93p-ca.crt
/Users/marian/.config/gsctl/certs/0r93p-7f3f88dafe-client.crt
/Users/marian/.config/gsctl/certs/0r93p-7f3f88dafe-client.key
Switched to kubectl context 'giantswarm-0r93p'

kubectl is set up. Check it using this command:

    kubectl cluster-info

Whenever you want to switch to using this context:

    kubectl config use-context giantswarm-0r93p

```

```nohighlight
$ openssl x509 -in /Users/marian/.config/gsctl/certs/0r93p-7f3f88dafe-client.crt -noout -enddate
notAfter=Apr 20 15:06:49 2018 GMT
```

```nohighlight
$ gsctl create kubeconfig -c 0r93p --ttl 10
The value passed with --ttl is invalid.
Please provide a number and a unit, e. g. '10h', '1d', '1w'.
```

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -44,7 +44,7 @@ var (
 	cmdDescription string
 
 	// TTL (time to live) flag
-	cmdTTLDays int
+	cmdTTL string
 
 	// force flag. if set, no prompt should be displayed.
 	cmdForce bool
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -240,7 +240,7 @@ func handleCommonErrors(err error) {
 		subtext += "please try again in a few moments."
 	case IsUnknownError(err):
 		headline = "An error occurred."
-		subtext = "Please notify the Giant Swarm support team, or try listing releases again in a few moments.\n"
+		subtext = "Please notify the Giant Swarm support team, or try the command again in a few moments.\n"
 		subtext += fmt.Sprintf("Details: %s", err.Error())
 	default:
 		return
--- a/commands/create_keypair.go
+++ b/commands/create_keypair.go
@@ -60,7 +60,7 @@ func defaultCreateKeypairArguments() createKeypairArguments {
 		clusterID:                cmdClusterID,
 		commonNamePrefix:         cmdCNPrefix,
 		description:              description,
-		ttlHours:                 int32(cmdTTLDays) * 24,
+		ttlHours:                 int32(cmdTTL) * 24,
 	}
 }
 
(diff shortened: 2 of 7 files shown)


commands/commands.go
```diff
@@ -240,7 +240,7 @@ func handleCommonErrors(err error) {
 		subtext += "please try again in a few moments."
 	case IsUnknownError(err):
 		headline = "An error occurred."
-		subtext = "Please notify the Giant Swarm support team, or try listing releases again in a few moments.\n"
+		subtext = "Please notify the Giant Swarm support team, or try the command again in a few moments.\n"
```
marians: FYI: Unrelated boyscouting

commands/create_keypair.go
```diff
@@ -82,15 +87,25 @@ func init() {
 	CreateKeypairCommand.Flags().StringVarP(&cmdDescription, "description", "d", "", "Description for the key pair")
 	CreateKeypairCommand.Flags().StringVarP(&cmdCNPrefix, "cn-prefix", "", "", "The common name prefix for the issued certificates 'CN' field.")
 	CreateKeypairCommand.Flags().StringVarP(&cmdCertificateOrganizations, "certificate-organizations", "", "", "A comma separated list of organizations for the issued certificates 'O' fields.")
-	CreateKeypairCommand.Flags().IntVarP(&cmdTTLDays, "ttl", "", 30, "Duration until expiry of the created key pair in days")
+	CreateKeypairCommand.Flags().StringVarP(&cmdTTL, "ttl", "", "30d", "Lifetime of the created key pair, e.g. 3h. Default: 30d (30 days).")
```
pipo02mix: What about enumerate the units allowed (h, w, d, y)

marians: Good point. Added that, and removed the duplicate info on the default. cobra adds that automatically.

commands/create_keypair.go
```diff
@@ -132,7 +147,8 @@ func verifyCreateKeypairPreconditions(args createKeypairArguments) error {
 }
 
 func createKeyPairRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
-	args := defaultCreateKeypairArguments()
+	args, _ := defaultCreateKeypairArguments()
+
```
pipo02mix: why here err is not checked?

util/duration_test.go
```diff
@@ -26,10 +29,32 @@ var durationTest = []struct {
 }
```
pipo02mix: Why do you say it is a breaking change if integers still working?

commands/create_keypair.go
marians: Because this is the second time `defaultCreateKeypairArguments()` is called. If errors occur, that should happen in the first call.

This isn't great design. And deserves rethinking, but I'd rather not do that now, because it affects almost all commands.

util/duration_test.go
marians: You are probably looking at the test for a different function. The tests for parsing the user-provided strings from `--ttl` to are in `parseDurationTest` and `TestParseDuration`.

pipo02mix (approved):
LGTM

marians:
Thanks!

Issue #226, opened 2018-04-18
'list keypairs' should allow showing the full CN string of a key pair


marians:
Belongs to https://github.com/giantswarm/giantswarm/issues/2903

Currently several columns of the `gsctl list keypairs` output are truncated to make the table fit a relatively narrow terminal.

Example:

```nohighlight
CREATED                 EXPIRES                 ID          DESCRIPTION                                                       CN                      O
2018 Mar 20, 10:41 UTC  2018 Apr 19, 10:41 UTC  406b120c0…  Added by user user@giantswarm.io using 'gsctl create kubeconfig'  user@giantswarm…        system:masters
2018 Mar 27, 12:30 UTC  2018 Apr 26, 12:30 UTC  02b9fad3c…  Marian's key pair                                                 user@giantswarm.io.us…  system:masters
2018 Apr 09, 14:05 UTC  2018 May 09, 14:05 UTC  4f16e3ac4…  Added by user user@giantswarm.io using 'gsctl create kubeconfig'  user@giantswarm.io.us…  system:masters
2018 Apr 10, 08:24 UTC  2018 May 10, 08:24 UTC  4ce4d5469…  Added by user user@giantswarm.io using 'gsctl create kubeconfig'  user@giantswarm.io.us…  system:masters
2018 Apr 12, 13:19 UTC  2018 May 12, 13:19 UTC  05e4b8b44…  Added by user user@giantswarm.io using 'gsctl create kubeconfig'  user@giantswarm.io.us…  system:masters
```

We should offer a way to optionally display all table output as a whole, without truncation.

PR #230, opened 2018-04-24
Add switch for truncation

marians:
Fixes https://github.com/giantswarm/gsctl/issues/226

### DIFF
--- a/util/string.go
+++ b/util/string.go
@@ -2,9 +2,9 @@ package util
 
 import "strings"
 
-// Truncate takes a string and truncates it
-func Truncate(str string, length int) string {
-	if len(str) < length {
+// Truncate takes a string and truncates it (if doTruncate is true)
+func Truncate(str string, length int, doTruncate bool) string {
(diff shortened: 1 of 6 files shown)


util/string.go
```diff
@@ -2,9 +2,9 @@ package util
 
 import "strings"
 
-// Truncate takes a string and truncates it
-func Truncate(str string, length int) string {
-	if len(str) < length {
+// Truncate takes a string and truncates it (if doTruncate is true)
+func Truncate(str string, length int, doTruncate bool) string {
```
oponder: A bit weird to give a function a true/false boolean on wether or not to do what its name says it will do.

But I see how it means less conditionals in `list_keypairs.go`

oponder (approved):

util/string.go
marians: Right. Doesn't look nice, but helps keep the output code readable.

marians:
Thanks!

PR #234, opened 2018-05-08
show cluster: Reduce memory and storage precision to 2 digits

marians:
For https://github.com/giantswarm/giantswarm/issues/2848

This PR sets the precision of the GB values in `gsctl show cluster` to 2 digits.

This also means that values like `40` will now show up as `40.00`.

#### Before

```
gsctl show cluster 6ur24
...
CPU cores in workers:          10
RAM in worker nodes (GB):      40.960003
Storage in worker nodes (GB):  81.920006
```

#### After

```
gsctl show cluster 6ur24
...
CPU cores in workers:          10
RAM in worker nodes (GB):      40.96
Storage in worker nodes (GB):  81.92
```


pipo02mix (approved):
LGMT

Issue #236, opened 2018-05-28
gsctl/gsclientgen should implement v4 login function


marians:
As https://github.com/giantswarm/api-spec/pull/49 is now implemented, we should update gsclientgen to pick up the new function.

marians:
Depends on https://github.com/giantswarm/gsclientgen/issues/32

marians:
- Login was migrated to v4 in https://github.com/giantswarm/gsctl/pull/257
- Logout was migrated in https://github.com/giantswarm/gsctl/pull/251

PR #237, opened 2018-05-28
Use v4 functions for login and logout in gsctl

marians:
For https://github.com/giantswarm/giantswarm/issues/1524 https://github.com/giantswarm/gsctl/issues/236

Depends on  https://github.com/giantswarm/gsclientgen/pull/30

TODO:

- Fix tests
- Change gsclientgen back to `master`


Gopkg.toml
```diff
@@ -38,16 +38,12 @@
...
-  version = "0.7.1"
-
 [[constraint]]
   name = "github.com/giantswarm/columnize"
   version = "=2.0.2"
 
 [[constraint]]
-  branch = "master"
+  branch = "v4-login"
   name = "github.com/giantswarm/gsclientgen"
```
pipo02mix: should it be changed back to master after test?

marians: Yes, as it says in the main comment.

marians:
Sorry, I asked for a review too early. Seems I have to fix some unit test(s) first.

marians:
There is no point in doing this with the client generated by swagger-codegen. Reason: We have no access to the response body in the case of HTTP response status >= 400. And we need the response to know what went wrong in logins.

Closing. Will redo one we are generating the client using go-swagger.

PR #239, opened 2018-06-04
Use client generated by go-swagger

marians:
WIP


marians:
Will try this again at some point, after we have a client wrapper.

PR #240, opened 2018-06-13
Support the bearer scheme (SSO)

oponder:
Let gsctl support the alternative Bearer scheme for passing a JWT token to the API.

This adds the `--auth-scheme` optional global parameter to gsctl, which will let users effectively change the Authorization header that gsctl's api client uses.

This is part of multiple steps towards supporting SSO in gsctl.

```
$ ./gsctl list releases --help
Prints detail on all available releases.

A release is a software bundle that constitutes a cluster. It is identified by its semantic version number.

Usage:
  gsctl list releases [flags]

Flags:
  -h, --help   help for releases

Global Flags:
      --auth-scheme string   Authorization scheme to use (giantswarm or Bearer, case sensitive) (default "giantswarm")
      --auth-token string    Authorization token to use
      --config-dir string    Configuration directory path to use (default "/Users/oponder/.config/gsctl")
  -e, --endpoint string      The API endpoint to use
  -v, --verbose              Print more information
```

Related to: https://github.com/giantswarm/giantswarm/issues/2480

### DIFF
--- a/commands/create_kubeconfig.go
+++ b/commands/create_kubeconfig.go
@@ -70,6 +70,7 @@ const (
 // function and to the validation function
 type createKubeconfigArguments struct {
 	apiEndpoint       string
+	scheme            string
 	authToken         string
 	clusterID         string
 	description       string
@@ -86,6 +87,7 @@ type createKubeconfigArguments struct {
 func defaultCreateKubeconfigArguments() (createKubeconfigArguments, error) {
 	endpoint := config.Config.ChooseEndpoint(cmdAPIEndpoint)
 	token := config.Config.ChooseToken(endpoint, cmdToken)
+	scheme := config.Config.ChooseScheme(endpoint, cmdScheme)
 
 	description := cmdDescription
 	if description == "" {
@@ -104,6 +106,7 @@ func defaultCreateKubeconfigArguments() (createKubeconfigArguments, error) {
 
 	return createKubeconfigArguments{
 		apiEndpoint:       endpoint,
+		scheme:            scheme,
 		authToken:         token,
 		clusterID:         cmdClusterID,
 		description:       description,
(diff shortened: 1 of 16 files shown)


oponder:
The plan next is to add a flag to `gsctl login` or an separate command (`gsctl login-sso` ?) which will assist in acquiring a bearer token. 

When the token is acquired, it will be stored in the config in the same way, but now with the `Bearer` scheme instead of the `giantswarm` scheme.

When the `Bearer` scheme is being used for an endpoint, `gsctl` will also do a pre-check to see if the token has expired or not, and attempt to silently acquire a new one.

puja108 (approved):
so this flag is only for when you manually give an auth token or would it also be of use somewhere else?

oponder:
Yeah I don't expect anyone to use the new --auth-scheme flag on a normal basis, unless they're trying to test an existing jwt token they got from somewhere else. 

The normal use case will be to `sso-login` and then the config file takes care of things behind the scenes.

puja108:
I would make it `sso` instead of `sso-login` as latter would be kinda duplicating the meaning (`sign on` and `login`)

commands/create_kubeconfig.go
```diff
@@ -454,7 +457,7 @@ func createKubeconfig(args createKubeconfigArguments) (createKubeconfigResult, e
 			Kind:           "Config",
 			CurrentContext: args.contextName,
 			Clusters: []KubeconfigNamedCluster{
-				KubeconfigNamedCluster{
```
tuommaki: <3

tuommaki (approved):
In general this looks good.  Couple things I noticed for the future improvement (not to be done here, but separately when there's suitable time for it):
- Many structs don't have alphabetical ordering of fields. It kind of glares out when reading the code when we mostly align fields pretty well :)
- It could possibly make sense to refactor `gsctl` in general so that auth scheme and token would be tightly coupled in a struct and handled as a pair in code as they are effectively very tightly coupled in the real world (it's not possible to use `giantswarm` scheme and `bearer` token etc., but this can be a bit bigger change so probably not worth to spend time right now.

oponder:
@tuommaki thanks for the review!

The refactorings you mention are definitely on my agenda now. I was already playing around with the idea of the struct for the auth information, and after doing a bit more exploratory coding in: https://github.com/giantswarm/gsctl/pull/241 I find that I would benefit from that a lot. So I'll work on that sooner rather than later.

I also want to extract the client into a separate concern somehow. It should be possible to extract the setting of the scheme and token for each command run, to avoid having to change multiple files.

PR #241, opened 2018-06-21
Add `sso` command for fetching a access token from auth0.

oponder:
This adds the `gsctl sso` command which will authenticate at auth0 and store the users access token.

Access tokens last only for 5 minutes, this PR does not yet deal with using a refresh token to get new access tokens silently.

~I will go back to my "Support Bearer Scheme" branch and do a few more refactorings (or possibly merge that and then do more refactorings in a new PR) before returning here to finish this.~

Related to: https://github.com/giantswarm/giantswarm/issues/2480


oponder:
While the code changes are quite exciting, running the command is not very eventful.

All you gotta do is build this branch, and run `./gsctl sso`. Your browser should open up with Auth0's SSO interface, and things should hopefully be intuitive from there.

1. Checkout this `sso` branch of gsctl
2. run `go build`
3. pick an endpoint, for example ginger: `./gsctl select endpoint ginger`
4. run `./gsctl sso`, follow instructions
5. see that you can list organizations: `./gsctl list organizations`
6. Done!

puja108 (approved):
LGTMIIWFY

oponder:
@marians would appreciate your approval and UX eyes as well before merging.

marians:
First of all, triple kudos for this!!!

Some comments:

1. I am all for having this functionality appear in `gsctl login` and not be a dedicated command. Let's discuss how the exact syntax should look like.

2. For the new config attribute `scheme`, can we think of a more specific name? E. g. `auth_scheme`?

3. The `scheme` attribute is persisted to the config YAML even when empty. This could be avoided. Also after logout, the value `scheme: Bearer` is no longer needed in the config.

Out of scope of this PR, but at some point we'll probably want to add the `scheme` value to the output of `gsctl info -v`, where we already show the token.

marians:
Please let's also find a way to have `pkce/pkce.go` be unit-tested.

marians (changes requested):
See my comments above

oponder:
Alright!

- moved the command to: `gsctl login --sso`.

- Let the user know that `--password` or `--auth-token` in combination with `--sso` is a pointless combination

- Made sure the `scheme` attribute is persisted as `auth_scheme`, and not persisted at all when it is empty.

- Avoided calling the giantswarm api when logging out a jwt token, since there is no need to call the api in that case. Just delete the token from the config.

PTALA @marians

oponder:
Do you still have the notes from our talk yesterday? in case I missed something.

marians:
Checked my notes. Nothing more in there.

commands/login.go
```diff
@@ -41,279 +34,34 @@ The password has to be entered interactively or given as -p / --password flag.
...
-		apiEndpoint: config.Config.ChooseEndpoint(cmdAPIEndpoint),
-		email:       cmdEmail,
-		password:    cmdPassword,
-		verbose:     cmdVerbose,
-	}
-}
-
 func init() {
 	LoginCommand.Flags().StringVarP(&cmdPassword, "password", "p", "", "Password. If not given, will be prompted interactively.")
+	LoginCommand.Flags().BoolVarP(&cmdSSO, "sso", "", false, "Authenticate using Single Sign On through our identity provider.")
```
marians: We wanted this hidden, so we also need

    LoginCommand.Flags().MarkHidden("sso")

here

commands/login_sso.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/fatih/color"
+	"github.com/giantswarm/gsctl/client"
+	"github.com/giantswarm/gsctl/config"
+	"github.com/giantswarm/gsctl/pkce"
+	"github.com/spf13/cobra"
+)
+
+const (
+	activityName = "sso"
```
marians: Seems like this is unused.

IMO we can use the same activity name as for the standard login, so simply get rid of this constant.

commands/root.go
```diff
@@ -20,7 +20,7 @@ func init() {
 
 	RootCommand.PersistentFlags().StringVarP(&cmdAPIEndpoint, "endpoint", "e", "", "The API endpoint to use")
 	RootCommand.PersistentFlags().StringVarP(&cmdToken, "auth-token", "", "", "Authorization token to use")
-	RootCommand.PersistentFlags().StringVarP(&cmdScheme, "auth-scheme", "", "giantswarm", "Authorization scheme to use (giantswarm or Bearer, case sensitive)")
+	RootCommand.PersistentFlags().StringVarP(&cmdScheme, "auth-scheme", "", "", "Authorization scheme to use (giantswarm or Bearer, case sensitive)")
```
marians: Why do we need this flag? After our talk I thought we wouldn't need it.

marians (changes requested):

oponder:
Ok! realized this required a bit more work on the `ChooseScheme` function to ensure that `--auth-token` would still work when `Bearer` was set as the scheme in the config.yaml.

commands/login.go
oponder: Ok done!

marians:
There is one more thing I am not super happy with. If you check out https://github.com/giantswarm/gsctl/blob/master/docs/Command-Blueprint.md, you see that a cobra.Command should be set up like this:

```
PreRun:  verbNounPreRunOutput,
Run:     verbNounRunOutput,
```

I made an effort to have all commands work that way. Now login is an exception.

Could we make it again so that `loginPreRunOutput` calls the validation and in case of any errors prints info and exits? This should include the flag combination check of `--sso` with `--password` or `--auth-token`. And `loginRunOutput` should call `login` and print the output, no matter what mechanics are used. Within the `login` function we could call different functions based on whether SSO is used or not.

oponder:
I didn't want to duplicate the `if cmdSSO` branch any further than needed.

If we split it up like this, we'll have to check in more than one place on `if cmdSSO`. Unless I misunderstood how to achieve it.

Will push a commit to show you the result

oponder:
And it's resulting in some nested ifs and a if/else, which is a bit un go-like...

oponder:
Hmm... I'm having a hard time seeing the benefit.

Let me upload what i have now. I lost some nice error output, since I have to make errors for each of those cases and then handle them in that one spot.

Agreed that now it all flows through `login` and uses `loginResult` to communicate the whole process whether its for `sso` or for a `giantswarm` token. But I don't think the code got any better for it.

marians:
There is maybe only one benefit: That whenever working on a command, you don't have to wonder about the high level flow. This will become more obvious once you work on gsctl more ;-)

oponder:
@marians do you like where this last commit is going? if so I'll finish it up by making some special error cases for the various errors that can happen in loginSSO.

marians (approved):

PR #251, opened 2018-07-09
Migrate to new client: Logout

marians:
For https://github.com/giantswarm/gsctl/issues/236

This is the second of a row of PRs that transition to the use of a new client version. Depends on https://github.com/giantswarm/gsclientgen/pull/31

### TODO
- [x] Set gsclientgen branch to `master` once https://github.com/giantswarm/gsclientgen/pull/31 is merged
- [x] test timeouts for the new client

### Nice to have
- [x] Have a way to pass common (optional) request arguments, like additional headers, non-default timeout etc.
- [x] Common way to handle a set of standard errors (403, no connection, DNS problem, timeout).
  - [ ] ~~extend `handleCommonErrors` to handle new error type~~ - will have to be done in future PRs


Gopkg.toml
```diff
@@ -46,6 +46,10 @@
   name = "github.com/giantswarm/columnize"
   version = "=2.0.2"
 
+[[constraint]]
+  branch = "goswagger"
```
TheoBrigitte: Reminder to revert to master

client/client.go
```diff
@@ -2,69 +2,76 @@ package client
...
 
 	requestIDHeader string
 	cmdLine         string
 )
 
-// Configuration is the configuration to be used when creating a new API client
+func init() {
+	rand.Seed(time.Now().UnixNano())
+	requestIDHeader = randomRequestID()
+	cmdLine = getCommandLine()
```
TheoBrigitte: I would rather have `requestIDHeader` and `cmdLine` initialized in `NewV2` and possibly attached to `WrapperV2` as they belong to the request being made (e.g. `UserLogin`) rather than the package.

client/client.go
```diff
@@ -2,69 +2,76 @@ package client
...
+}
+
+// Configuration is the configuration to be used both by the latest as well
+// as the old client based on gsclientgen.v1
```
TheoBrigitte: General nitpick about documentation: it ends with a dot.

client/client.go
```diff
@@ -2,69 +2,76 @@ package client
...
-	if clientConfig.Timeout != 0 {
-		configuration.Timeout = &clientConfig.Timeout
+	u, err := url.Parse(conf.Endpoint)
+	if err != nil {
+		return nil, microerror.Mask(endpointInvalidError)
 	}
 
-	// set up client TLS so that custom CAs are accepted.
 	tlsConfig := &tls.Config{}
 	rootCertsErr := rootcerts.ConfigureTLS(tlsConfig, &rootcerts.Config{
```
TheoBrigitte: `rootCertsErr` should probably be `err`

client/client.go
```diff
@@ -145,54 +154,68 @@ func redactPasswordArgs(args []string) []string {
...
-}
-func (c *ClientWrapper) UserLogin(email string, body gsclientgen.LoginBodyModel, activityname string) (*gsclientgen.LoginResponseModel, *gsclientgen.APIResponse, error) {
-	response, apiResponse, err := c.client.UserLogin(email, body, c.requestID, activityname, cmdLine)
-	return response, apiResponse, err
+	response, err := w.gsclient.AuthTokens.CreateAuthToken(params, nil)
+	if err != nil {
+		return nil, clienterror.New(err)
+	}
+
+	return response.Payload, nil
```
TheoBrigitte: Why not returning the whole response here?
I feel it could be helpful in the future to be able to access the whole response object

client/client_test.go
```diff
@@ -98,3 +102,316 @@ func TestRedactPasswordArgs(t *testing.T) {
...
+
+	clientConfig.Endpoint = ts.URL
+
+	gsClient, err := NewV2(clientConfig)
+	if err != nil {
+		t.Error(err)
+	}
+
+	// just issue a request, don't care about the result
+	_, _ = gsClient.CreateAuthToken("email", "password", nil)
```
TheoBrigitte: When result is not to be read from drop the `=`
Just make it `gsClient.CreateAuthToken("email", "password", nil)`

client/client_test.go
```diff
@@ -98,3 +102,316 @@ func TestRedactPasswordArgs(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	ap := gsClient.DefaultAuxiliaryParams()
+	ap.RequestID = "request-id"
+	ap.CommandLine = "command-line"
+	ap.ActivityName = "activity-name"
+
+	_, _ = gsClient.CreateAuthToken("foo", "bar", ap)
```
TheoBrigitte: Just make it `gsClient.CreateAuthToken("foo", "bar", ap)`

client/client_test.go
```diff
@@ -98,3 +102,316 @@ func TestRedactPasswordArgs(t *testing.T) {
...
+
+	responseBody, err := gsClient.DeleteAuthToken("test-token", nil)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if responseBody.Code != "RESOURCE_DELETED" {
+		t.Errorf("Didn't get the RESOURCE_DELETED message. Got '%s'", responseBody.Code)
+	}
+}
```
TheoBrigitte: Good job for getting those tests :+1:

client/clienterror/clienterror.go
```diff
@@ -0,0 +1,173 @@
...
+	"net/url"
+	"reflect"
+
+	"github.com/go-openapi/runtime"
+
+	"github.com/giantswarm/gsclientgen/client/auth_tokens"
+)
+
+// APIError is our structure to carry all error information we care about
+// from the api client to the CLI
```
TheoBrigitte: dot ending reminder

client/clienterror/clienterror.go
```diff
@@ -0,0 +1,173 @@
...
+				return ae
+			}
+		}
+
+		return ae
+	}
+
+	// Timeout / context deadline exceeded
+	// Note: We'd love to do type assertion here, but there is no exported type to assert with.
+	errorType := reflect.TypeOf(err)
```
TheoBrigitte: I would rather have `err == context.DeadlineExceeded`

commands/logout.go
```diff
@@ -94,30 +94,9 @@ func logout(args logoutArguments) error {
...
-	}
-
-	if logoutResponse.StatusCode != apischema.STATUS_CODE_RESOURCE_DELETED {
-		return microerror.Maskf(unspecifiedAPIError, "response: %v", logoutResponse)
-	}
+	ap := ClientV2.DefaultAuxiliaryParams()
+	ap.ActivityName = logoutActivityName
 
-	return nil
+	_, err := ClientV2.DeleteAuthToken(args.token, ap)
```
TheoBrigitte: It would be good to mask the error here in case there is any, it makes debugging easier by providing consistent stack trace where we can look up the code function by function. If we skip masking the error here, this function will not show in stack trace and there would be a missing step which feels wrong.

client/client.go
marians: Changed with the latest commit

client/client.go
marians: Added many with the last commit

client/client_test.go
marians: changed

client/client_test.go
marians: changed

client/clienterror/clienterror.go
marians: Cool! I missed that.

client/client.go
marians: Changed to returning the whole response

commands/logout.go
marians: Done

marians:
@TheoBrigitte Thanks for the great review! I think I addressed all comments. Could you check again and approve if possible?

client/client.go
TheoBrigitte: The thinking is here, is to provide as access to as much information as we can to the caller function and not having to change that in the future, because caller need access to `response.Something`.

marians:
gsclientgen branch reverted to master.

Gopkg.toml
marians: Done

PR #261, opened 2018-07-19
Migrate GetInfo calls to new client

marians:
This replaces the legacy client's GetInfo function with one using the new client.


r7vme (approved):
LGTM

PR #264, opened 2018-07-19
Migrate 'list releases' command to the new client

marians:
Ongoing client migration

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -15,6 +15,7 @@ import (
 	"github.com/giantswarm/gsclientgen/client/clusters"
 	"github.com/giantswarm/gsclientgen/client/info"
 	"github.com/giantswarm/gsclientgen/client/key_pairs"
+	"github.com/giantswarm/gsclientgen/client/releases"
 	"github.com/giantswarm/gsclientgen/models"
 	"github.com/giantswarm/gsctl/client/clienterror"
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -406,3 +407,48 @@ func (w *WrapperV2) GetInfo(p *AuxiliaryParams) (*info.GetInfoOK, error) {
...
+		if p.ActivityName != "" {
+			params.SetXGiantSwarmActivity(&p.ActivityName)
+		}
+		if p.CommandLine != "" {
+			params.SetXGiantSwarmCmdLine(&p.CommandLine)
+		}
+		if p.RequestID != "" {
+			params.SetXRequestID(&p.RequestID)
+		}
+	}
```
kopiczko: That's a lot of repetition here. It looks like there can be interface extracted and then we could apply those params in a generic way.

kopiczko (approved):

client/client.go
marians: +1! I really hope so.

kopiczko: Uncertain if we understood each other. This is in your reach. You can create this interface yourself. In Go you don't have to make struct implement interface it just has to have interface's methods.

marians: Yeah, I know. I just haven't taken the time to do it yet and it's something I am not used to doing.

PR #266, opened 2018-07-20
Migrate getCluster operation to new client

marians:


client/client.go
```diff
@@ -307,6 +307,26 @@ func (w *WrapperV2) DeleteCluster(clusterID string, p *AuxiliaryParams) (*cluste
 	return response, nil
 }
 
+// GetCluster fetches details on a clsuter using the latest client.
```
JosephSalisbury: `cluster`

client/client.go
```diff
@@ -307,6 +307,26 @@ func (w *WrapperV2) DeleteCluster(clusterID string, p *AuxiliaryParams) (*cluste
...
+func (w *WrapperV2) GetCluster(clusterID string, p *AuxiliaryParams) (*clusters.GetClusterOK, error) {
+	if w == nil {
```
JosephSalisbury: This feels weird. My gut says you shouldn't be able to get into a point where you're calling against nil

marians: Ideally no. But it did happen to me when the client wasn't initialized properly. So how could I prevent it in a more elegant way?

kopiczko: LoL https://play.golang.org/p/GoslHrX9Vtm
I didn't know that's possible

JosephSalisbury: It makes me worried - it feels like you have some cases where you're not constructing the client correctly. I'd not have this check, and solve it there.

TheoBrigitte: > I didn't know that's possible

I did the exact same thing (including the lol message)

There is also a common good practice that says, avoid pointer whenever possible and only use pointer receiving method when you actually need to change some value in the object.
So here you can change from `func (w *WrapperV2) GetCluster` to `func (w WrapperV2) GetCluster`

kopiczko: Where is the `WrapperV2` initialized? Why it can be `nil` and we should handle that?

marians:
Can this get a final approval please? Thx!

marians:
We'll change it to `func (w WrapperV2)` at some point.

PR #269, opened 2018-07-23
Migrate get user orgs

marians:
TODO:
- [ ] handle specific errors in clienterror

### DIFF
--- a/commands/list_organizations.go
+++ b/commands/list_organizations.go
@@ -1,7 +1,6 @@
 package commands
 
 import (
-	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -10,7 +9,9 @@ import (
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 
+	"github.com/giantswarm/gsctl/client/clienterror"
 	"github.com/giantswarm/gsctl/config"
+	"github.com/giantswarm/microerror"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


kopiczko:
I just noticed you requested review only from me. It would be good to request from @giantswarm/sig-operator for awareness.

commands/list_organizations.go
```diff
@@ -55,30 +88,38 @@ func listOrgs(cmd *cobra.Command, args []string) {
...
+		output = color.YellowString("No organizations available\n")
+	} else {
+		// sort orgs by Id
+		slice.Sort(response.Payload[:], func(i, j int) bool {
+			return response.Payload[i].ID < response.Payload[j].ID
+		})
+
+		output = color.CyanString("ORGANIZATION") + "\n"
+		for _, org := range response.Payload {
+			output = output + org.ID + "\n"
```
xh3b4sd: This looks like the table is rendered manually. Isn't there a lib we can use to do this?

commands/list_organizations.go
```diff
@@ -55,30 +88,38 @@ func listOrgs(cmd *cobra.Command, args []string) {
...
 		}
-		return output, nil
 	}
-	return "", APIError{fmt.Sprintf("Unhandled response code: %v", apiResponse.Response.StatusCode), *apiResponse}
```
xh3b4sd: The error handling here does not conform to what the dev team usually agrees on. See https://github.com/giantswarm/fmt/blob/master/go/errors.md.

marians: @xh3b4sd Looks like you are commenting on code that's being removed. I could explain the new error handling to you, if you're interested (line 97 to 197). Let me know.

PR #270, opened 2018-07-23
Modify wrapper receiver, remove checks for receiver being nil

marians:
This eliminates the need for nil checks in all wrapper functions.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -177,7 +177,7 @@ type AuxiliaryParams struct {
 
 // DefaultAuxiliaryParams returns a partially pre-populated AuxiliaryParams
 // object.
-func (w *WrapperV2) DefaultAuxiliaryParams() *AuxiliaryParams {
+func (w WrapperV2) DefaultAuxiliaryParams() *AuxiliaryParams {
 	return &AuxiliaryParams{
 		CommandLine: getCommandLine(),
 		RequestID:   randomRequestID(),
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -251,16 +251,12 @@ func setParamsWithAuthorization(p *AuxiliaryParams, w *WrapperV2, params paramSe
 }
 
 // CreateAuthToken creates an auth token using the latest client.
-func (w *WrapperV2) CreateAuthToken(email, password string, p *AuxiliaryParams) (*auth_tokens.CreateAuthTokenOK, error) {
-	if w == nil {
-		return nil, microerror.Mask(clientV2NotInitializedError)
-	}
-
+func (w WrapperV2) CreateAuthToken(email, password string, p *AuxiliaryParams) (*auth_tokens.CreateAuthTokenOK, error) {
```
kopiczko: I think pointer receiver is fine. Especially when you will need to e.g. use a mutex.
I just don't understand why this could ever be nil and would rather look for the root cause of the problem. Then just remove these nil checks.

marians: The root cause was that the client was not initialized correctly, when writing a unit test. If we think this is not dangerous, as it won't happen at normal runtime, then I'm fine with leaving out the nil checks.

marians: Now changed back to pointer receiver.

kopiczko: Yes I think unit test should be just fixed. No need to handle that.

PR #275, opened 2018-07-27
Handle keypair exceeding limits

marians:
For https://github.com/giantswarm/giantswarm/issues/3731

This PR does two things:

- Adds handling of a bad request error with a hint to adjust the TTL. Unfortunately the API doesn't give a specific error currently, so we can't be more specific than that.

- Adds validation for the TTL time being above 292 years, which is maybe a bit silly, but it's the absolute maximum that can be represented using a Go `time.Duration`.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -138,8 +138,7 @@ func handleCommonErrors(err error) {
 	var subtext = ""
 
 	// V2 client error handling
-	convertedErr, ok := err.(*clienterror.APIError)
-	if ok {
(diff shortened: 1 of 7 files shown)


calvix:
AFAIK vault has max-lease-ttl argument which decide what is that max-ttl for a certificate that can be issued. I  don't know what is current value, could be something around year.

pipo02mix (approved):
LGTMIIWFY

PR #277, opened 2018-08-01
Update ping command to use / instead of /v1/ping

marians:
In addition, the output is changed to a meaningful precision.

Before:

    Ping took 165.135135ms

After:

    Ping took 113 Milliseconds

### DIFF
--- a/commands/ping.go
+++ b/commands/ping.go
@@ -46,7 +46,7 @@ func runPingCommand(cmd *cobra.Command, args []string) {
 	}
 
 	fmt.Println(color.GreenString("API connection is fine"))
-	fmt.Printf("Ping took %d Milliseconds\n", int(duration.Seconds()*1000))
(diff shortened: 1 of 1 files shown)


commands/ping.go
```diff
@@ -46,20 +46,20 @@ func runPingCommand(cmd *cobra.Command, args []string) {
 	}
 
 	fmt.Println(color.GreenString("API connection is fine"))
-	fmt.Printf("Ping took %v\n", duration)
+	fmt.Printf("Ping took %d Milliseconds\n", int(duration.Seconds()*1000))
```
TheoBrigitte: This could be simplify as 
`fmt.Printf("Ping took %d Milliseconds\n", duration / time.Millisecond)`

TheoBrigitte (changes requested):

marians:
PTAL @TheoBrigitte

JosephSalisbury (approved):

PR #278, opened 2018-08-01
SSO login: add verbose output, handle connection error

marians:
This originated in user feedback regarding a problem with `gsctl login --sso`. The root cause was VPN issues.

With the new client, a bad VPN connection or IP-based blocking from the API results in an error that originates from the attempt to parse the response body as JSON. In this case we don't even get the HTTP status response code back, so all we have to test against is the error string.

While at it, I added optional output to be activated using `-v|--verbose` in case of an error.

### DIFF
--- a/client/clienterror/clienterror.go
+++ b/client/clienterror/clienterror.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/go-openapi/runtime"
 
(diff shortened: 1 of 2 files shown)


client/clienterror/clienterror.go
```diff
@@ -329,7 +330,21 @@ func New(err error) *APIError {
 		return ae
 	}
 
-	fmt.Printf("Error: %#v\n", err)
```
marians: This is the removal of a left-over.

pipo02mix (approved):
LGTM :)

PR #280, opened 2018-08-06
Introduce AuthHeaderGetter for eventual Access Token refreshing.

oponder:
This PR introduces my mechanism for being able to run some code on each client run.

Instead of passing it the auth header to use, we pass it a function it can use to get the auth header.

This function will then in the next PR be further developed to check whether or not the token we have
is valid, and if not, it will refresh the token.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -10,6 +10,10 @@ import (
 	"strings"
 	"time"
 
+	httptransport "github.com/go-openapi/runtime/client"
+	"github.com/go-openapi/strfmt"
+	rootcerts "github.com/hashicorp/go-rootcerts"
+
 	gsclient "github.com/giantswarm/gsclientgen/client"
 	"github.com/giantswarm/gsclientgen/client/auth_tokens"
 	"github.com/giantswarm/gsclientgen/client/clusters"
@@ -18,11 +22,9 @@ import (
 	"github.com/giantswarm/gsclientgen/client/organizations"
 	"github.com/giantswarm/gsclientgen/client/releases"
 	"github.com/giantswarm/gsclientgen/models"
-	"github.com/giantswarm/gsctl/client/clienterror"
 	"github.com/giantswarm/microerror"
-	httptransport "github.com/go-openapi/runtime/client"
-	"github.com/go-openapi/strfmt"
-	rootcerts "github.com/hashicorp/go-rootcerts"
+
+	"github.com/giantswarm/gsctl/client/clienterror"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -41,8 +43,8 @@ func init() {
 
 // Configuration is the client configuration
 type Configuration struct {
-	// AuthHeader is the header we should use to make API calls.
```
oponder: This is the main story of this PR.

We replace `AuthHeader string` with a `AuthHeaderGetter` that returns a string.

That gives us an opportunity to run checks against the AuthHeader before it gets used.

config/config.go
```diff
@@ -324,13 +334,32 @@ func (c *configStruct) Logout(endpointURL string) {
...
 
+// AuthHeaderGetter returns a function that can get the auth header for a given endpoint that the client can use.
+// The returned function will attempt to refresh the token in case the scheme is Bearer and the token is expired.
+func (c *configStruct) AuthHeaderGetter(endpoint string, overridingToken string) func() (authheader string, err error) {
+	return func() (string, error) {
+		token := c.ChooseToken(endpoint, overridingToken)
+		scheme := c.ChooseScheme(endpoint, overridingToken)
+
+		// If the scheme is Bearer, first verify that the token is expired.
+		// And if it is expired, then try to refresh it.
```
oponder: The next PR would then implement what this comment is saying. But first I'd like to get approval on the approach taken in this PR.

client/client.go
```diff
@@ -74,6 +76,9 @@ type WrapperV2 struct {
 
 // NewV2 creates a client based on the latest gsclientgen version.
 func NewV2(conf *Configuration) (*WrapperV2, error) {
+	if conf.AuthHeaderGetter == nil {
+		conf.AuthHeaderGetter = func() (string, error) { return "", nil }
+	}
```
marians: In what situations will returning an empty string provide useful? Is that for unauthenticaded access of something?

marians: ... probably for the `CreateAuthToken` function?

oponder: This is really just in case someone does not provide a Getter during initialization of this client. (The tests for example do this)

I could throw an error somehow and make it a required configuration parameter. But at least in this case we don't get a panic when it tries to run a non existing function.

I could instead throw an error, but this error would only show up when gsctl is run, and not at compile time.

oponder: > .. probably for the CreateAuthToken function?

Nope, the `CreateAuthToken` function does not use `SetParamsWithAuthorization` so it wont end up calling the `AuthHeaderGetter` either.

marians:
LGTMIIWFY

PR #283, opened 2018-08-08
Print "No clusters available" when no clusters found with 'gsctl list clusters'

marians:
`gsctl list clusters` would, as the only command, not print anything when it successfully communicated with the API but didn't get any items back to print. Changing this to match other list commands, as it is really unexpected.

### DIFF
--- a/commands/list_clusters.go
+++ b/commands/list_clusters.go
@@ -120,7 +120,7 @@ func clustersTable(args listClustersArguments) (string, error) {
 	}
 
 	if len(response.Payload) == 0 {
-		return color.YellowString("No clusters available"), nil
+		return color.YellowString("No clusters"), nil
 	}
 	// table headers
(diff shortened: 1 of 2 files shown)


commands/list_clusters.go
```diff
@@ -120,7 +120,7 @@ func clustersTable(args listClustersArguments) (string, error) {
 	}
 
 	if len(response.Payload) == 0 {
-		return "", nil
+		return color.YellowString("No clusters available"), nil
```
JosephSalisbury: `available` feels like the wrong term here - makes me think that clusters do exist, but aren't accessible. Is `No clusters` enough?

r7vme: or `No clusters exist`

r7vme (approved):
LGTM after workind comments addressed

JosephSalisbury (approved):

PR #285, opened 2018-08-09
Validate and renew access token if needed.

oponder:
This PR implements the final step of the refresh token story.

It checks if the token we're about to use is valid. If it isn't it'll perform the steps needed to get a new one using the stored refresh token.

Towards: https://github.com/giantswarm/giantswarm/issues/2480

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -156,6 +156,10 @@ func handleCommonErrors(err error) {
 			headline = "Unauthorized"
 			subtext = "Something went wrong during a OIDC operation: " + err.Error() + "\n"
 			subtext += "Please try logging in again."
+		case oidc.IsRefreshError(err):
+			headline = "Unable to refresh your SSO token."
+			subtext = err.Error() + "\n"
+			subtext += "Please try loging in again using: gsctl login --sso"
 		case IsNotLoggedInError(err):
 			headline = "You are not logged in."
 			subtext = "Use 'gsctl login' to login or '--auth-token' to pass a valid auth token."
@@ -186,10 +190,6 @@ func handleCommonErrors(err error) {
 			headline = "The API didn't send a response."
 			subtext = "Please check your connection using 'gsctl ping'. If your connection is fine,\n"
 			subtext += "please try again in a few moments."
-		case oidc.IsRefreshError(err):
-			headline = "Unable to refresh your SSO token."
-			subtext = err.Error() + "\n"
-			subtext += "Please try loging in again using: gsctl login --sso"
 		case IsUnknownError(err):
 			headline = "An error occurred."
 			subtext = "Please notify the Giant Swarm support team, or try the command again in a few moments.\n"
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -17,6 +17,7 @@ import (
 	"github.com/giantswarm/gsctl/client"
 	"github.com/giantswarm/gsctl/client/clienterror"
 	"github.com/giantswarm/gsctl/config"
+	"github.com/giantswarm/gsctl/oidc"
 )
 
 var (
@@ -152,6 +153,10 @@ func handleCommonErrors(err error) {
 		case client.IsEndpointNotSpecifiedError(err):
 			headline = "No endpoint has been specified."
 			subtext = "Please use the '-e|--endpoint' flag."
+		case oidc.IsAuthorizationError(err):
+			headline = "Unauthorized"
+			subtext = "Something went wrong during a OIDC operation: " + err.Error() + "\n"
+			subtext += "Please try logging in again."
 		case IsNotLoggedInError(err):
 			headline = "You are not logged in."
 			subtext = "Use 'gsctl login' to login or '--auth-token' to pass a valid auth token."
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -16,7 +16,6 @@ import (
 
 	"github.com/giantswarm/gsctl/client"
 	"github.com/giantswarm/gsctl/client/clienterror"
-	"github.com/giantswarm/gsctl/config"
 	"github.com/giantswarm/gsctl/oidc"
 )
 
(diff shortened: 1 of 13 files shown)


commands/login.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/fatih/color"
 	"github.com/giantswarm/gsctl/client"
 	"github.com/giantswarm/gsctl/config"
-	"github.com/giantswarm/gsctl/pkce"
```
oponder: I did a refactoring in this PR also, renaming the `pkce` package to `oidc` since the Refresh Token behaviour I introduced here is not really `pkce`. So I lifted the concept a bit higher, now `oidc` is the package that knows about things like `pkce` and `refresh token`

marians:
Did you see the test output?

```
=== RUN   Test_CreateClusterSuccessfully
Requesting new cluster for organization 'acme'
--- FAIL: Test_CreateClusterSuccessfully (0.00s)
	create_cluster_test.go:148: Case 0: Minimal arguments
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x86e617]

goroutine 58 [running]:
testing.tRunner.func1(0xc4202325a0)
	/usr/local/go/src/testing/testing.go:742 +0x29d
panic(0x96b800, 0xd79b00)
	/usr/local/go/src/runtime/panic.go:502 +0x229
github.com/giantswarm/gsctl/config.(*configStruct).AuthHeaderGetter.func1(0x4128f8, 0x50, 0x9daea0, 0x1)
	/go/src/github.com/giantswarm/gsctl/config/config.go:353 +0x147
github.com/giantswarm/gsctl/client.setParamsWithAuthorization(0xc4200b8ac0, 0xc4200ae780, 0xa95800, 0xc4200d2640, 0xe, 0x70)
	/go/src/github.com/giantswarm/gsctl/client/client.go:250 +0x9d
github.com/giantswarm/gsctl/client.(*WrapperV2).CreateCluster(0xc4200ae780, 0xc4200d25f0, 0xc4200b8ac0, 0x1, 0x1, 0x2f)
	/go/src/github.com/giantswarm/gsctl/client/client.go:297 +0x8a
github.com/giantswarm/gsctl/commands.addCluster(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xa10b04, 0x4, ...)
	/go/src/github.com/giantswarm/gsctl/commands/create_cluster.go:507 +0x419
github.com/giantswarm/gsctl/commands.Test_CreateClusterSuccessfully(0xc4202325a0)
	/go/src/github.com/giantswarm/gsctl/commands/create_cluster_test.go:194 +0x5a2
testing.tRunner(0xc4202325a0, 0xa4ae28)
	/usr/local/go/src/testing/testing.go:777 +0xd0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:824 +0x2e0
```

oidc/refresh_token.go
```diff
@@ -0,0 +1,82 @@
...
+		refreshResponse.Error = "unknown_error"
+		refreshResponse.ErrorDescription = "Unable to construct POST request for Auth0."
+		return refreshResponse, microerror.Maskf(authorizationError, refreshResponse.Error)
+	}
+
+	req.Header.Add("content-type", "application/json")
+
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		refreshResponse.Error = "unknown_error"
```
marians: Why are we doing this underscore thing here with `unknown_error`?

oidc/refresh_token.go
```diff
@@ -0,0 +1,82 @@
...
+		refreshResponse.ErrorDescription = "Unable to perform POST request to Auth0."
+		return refreshResponse, microerror.Maskf(authorizationError, refreshResponse.Error)
+	}
+
+	defer res.Body.Close()
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		refreshResponse.Error = "unknown_error"
+		refreshResponse.ErrorDescription = "Got an unparseable error from Auth0. Possibly the Auth0 service is down. Try again later."
+		return refreshResponse, microerror.Maskf(authorizationError, refreshResponse.Error)
```
marians: We should return a specific error here, not `authorizationError`. In `handleCommonError` for `authorizationError` we print `Unauthorized` as a headline, which can be misleading. In this case here the authentication process failed, so we should allow ourselves to show a more appropriate message to the user.

oidc/refresh_token.go
```diff
@@ -0,0 +1,82 @@
...
+	// This is a real error from Auth0, in this case we have Error and ErrorDescription
+	// set by what Auth0 sent us.
+	if refreshResponse.Error != "" {
+		return refreshResponse, microerror.Maskf(authorizationError, refreshResponse.Error)
+	}
+
+	if res.StatusCode != 200 {
+		refreshResponse.Error = "unknown_error"
+		refreshResponse.ErrorDescription = "Got an unparseable error from Auth0. Possibly the Auth0 service is down. Try again later."
+		return refreshResponse, microerror.Maskf(authorizationError, refreshResponse.Error)
```
marians: See comment above. Please don't return `authorizationError`. Probably returning the same one as in Line 64 would make sense.

marians:
I tested and it works as expected! I'd like some more specific error handling though.

oponder:
> I tested and it works as expected! I'd like some more specific error handling though.

Thanks! Ok, I'll process that in a bit. Going out quickly to observe the singaporean national day, but will apply this feedback later today.

oponder:
The output now looks like this when something goes wrong:

```
$ go build && ./gsctl list clusters
Unable to refresh your SSO Token.
Unknown or invalid refresh token.: refresh error
Please try loging in again using: gsctl login --sso
```

commands/commands.go
```diff
@@ -182,9 +186,10 @@ func handleCommonErrors(err error) {
 			headline = "The API didn't send a response."
 			subtext = "Please check your connection using 'gsctl ping'. If your connection is fine,\n"
 			subtext += "please try again in a few moments."
-		case config.IsUnableToRefreshTokenErrorr(err):
-			headline = "Unable to refresh your SSO Token."
-			subtext = "Please try loging in again using: gsctl login --sso"
+		case oidc.IsRefreshError(err):
+			headline = "Unable to refresh your SSO token."
+			subtext = err.Error() + "\n"
+			subtext += "Please try loging in again using: gsctl login --sso"
```
marians: This could be placed immediately before or after with `case oidc.IsAuthorizationError(err):`. Minor.

PR #292, opened 2018-08-31
Add test for the cobra root command

marians:


marians:
Not valuable. Closing.

PR #295, opened 2018-09-10
improves order of arguments in command docs for more convenience

xh3b4sd:


commands/create_kubeconfig.go
```diff
@@ -42,11 +42,11 @@ Examples:
...
   gsctl create kubeconfig -c my0c3
 
-  gsctl create kubeconfig -c my0c3 --self-contained ./kubeconfig.yaml
+  gsctl create kubeconfig --self-contained ./kubeconfig.yaml -c my0c3
 
-  gsctl create kubeconfig -c my0c3 --ttl 3h -d "Key pair living for only 3 hours"
+  gsctl create kubeconfig --ttl 3h -d "Key pair living for only 3 hours" -c my0c3
 
-  gsctl create kubeconfig -c my0c3 --certificate-organizations system:masters
+  gsctl create kubeconfig --certificate-organizations system:masters -c my0c3
```
xh3b4sd: I want the cluster ID to be the last argument because my cursor is at the end of the line when searching my history and I need less keystrokes to reuse the history command for the current cluster I am working with that way.

marians (changes requested):
How do you think this improves the usage?

In fact I think it makes sense to have mandatory flags (here: `-c`) in the front always.

commands/create_kubeconfig.go
marians: I don't think that can be generalized as an improvement for all users.

I would suggest you to create a function/alias in your shell which just takes the argument of something like `gck <cluster-id>` and executes `gsctl create kubeconfig --certificate-organizations system:masters -c <cluster-id>` with it.

xh3b4sd: Disagree and commit.

PR #298, opened 2018-10-02
Debug token issued at time errors.

oponder:
This branch is here to help us debug Issued At Time validation errors. Which manifest as `Token created in the future?` in gsctl.

It doesn't need to be merged, but @corest @jgsqware or @indyfree could build it and use it for a while to gather some data.

### DIFF
--- a/oidc/jwt.go
+++ b/oidc/jwt.go
@@ -2,7 +2,9 @@ package oidc
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/cenkalti/backoff"
 	jwt "github.com/dgrijalva/jwt-go"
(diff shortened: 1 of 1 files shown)


marians:
Suggestion: Print the time difference in seconds and make it mergeable. It's probably most helpful if everyone with that problem uses this version.

oponder:
Ok, now it shows the difference in seconds. The IAT claim is a unix timestamp in seconds, no sub second resolution. 

```
$ go build && ./gsctl login --sso

Your browser should now be opening:
...

Issued at:  1538463850
Now:        1538463847
Difference: 3 seconds

Token created in the future?
It appears as if your system time is behind the actual time. Please adjust the time and make sure
that it is automatically synchronized with a time service. Otherwise SSO login does not work.
```

marians (approved):

Issue #190, opened 2018-01-23
gsctl list releases sorting should be reversed (newest last)


marians:
Currently it's oldest last, should be oldest first, to have the latest visible.

teemow:
Imo the list would also be much easier to read if it would come by default in a compact "table view". and then something like `gsctl show release --version 2.1.2` can list all the details of a particular release. Sorry for piggybacking this issue :)

teemow:
@marians are we good here with the new `gsctl list releases`?

marians:
Yes, thanks!

PR #300, opened 2018-10-03
change 'list release' to table and add 'show release'

teemow:
list release: displays a shorter list of releases now. there is now also
a status column that indicates active, future and past releases.

show release: displays the former yaml of the releases list but with a
single release only

### DIFF
--- a/commands/list_releases.go
+++ b/commands/list_releases.go
@@ -131,7 +131,7 @@ func listReleasesRunOutput(cmd *cobra.Command, extraArgs []string) {
 	major = 0
 	status = "deprecated"
 
-	for _, release := range releases {
+	for i, release := range releases {
 		created := util.ShortDate(util.ParseDate(*release.Timestamp))
 		kubernetes_version := "n/a"
 		container_linux_version := "n/a"
@@ -145,8 +145,17 @@ func listReleasesRunOutput(cmd *cobra.Command, extraArgs []string) {
 
 		if err == nil {
 			if version.Major() > major {
+				// Found new major release.
 				major = version.Major()
-				status = "deprecated"
+
+				// If this is a new major version and the last release
+				// likelihood is high that this is a wip release and
+				// not deprecated.
+				if i == len(releases)-1 {
+					status = "wip"
+				} else {
+					status = "deprecated"
+				}
 			}
 
 			if release.Active {
--- a/commands/list_releases.go
+++ b/commands/list_releases.go
@@ -129,7 +129,7 @@ func listReleasesRunOutput(cmd *cobra.Command, extraArgs []string) {
 	var major int64
 	var status string
 	major = 0
-	status = "-"
+	status = "deprecated"
 
 	for _, release := range releases {
 		created := util.ShortDate(util.ParseDate(*release.Timestamp))
--- a/commands/list_releases.go
+++ b/commands/list_releases.go
@@ -119,11 +119,11 @@ func listReleasesRunOutput(cmd *cobra.Command, extraArgs []string) {
 	output := []string{strings.Join([]string{
 		color.CyanString("VERSION"),
 		color.CyanString("STATUS"),
+		color.CyanString("CREATED"),
 		color.CyanString("KUBERNETES"),
 		color.CyanString("CONTAINERLINUX"),
 		color.CyanString("COREDNS"),
 		color.CyanString("CALICO"),
-		color.CyanString("CREATED"),
 	}, "|")}
 
 	var major int64
@@ -138,9 +138,9 @@ func listReleasesRunOutput(cmd *cobra.Command, extraArgs []string) {
 		coredns_version := "n/a"
 		calico_version := "n/a"
 
-		// as long as the status information is not specific in the API
+		// As long as the status information is not specific in the API
 		// we start with deprecated, find the active one and then switch
-		// to "wip" for each major version
+		// to "wip" for each major version.
 		version, err := semver.NewVersion(*release.Version)
 
 		if err == nil {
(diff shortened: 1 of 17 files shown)


teemow:
Example:

screen_2018-10-03-21-47-01

teemow:
related to https://github.com/giantswarm/gsctl/issues/190

commands/show_release.go
```diff
@@ -0,0 +1,173 @@
...
+		var headline = ""
+		var subtext = ""
+
+		// TODO: handle common and specific errors
+		switch {
+		case err.Error() == "":
+			return
+		default:
+			headline = err.Error()
+		}
```
marians: As the TODO marker indicates, we are missing error handling here for `releaseNotFoundError`.

I think it's also worthehile to use handleCommonErrors(err) here, too. But not sure. Maybe check how this is done in other commands.

commands/list_releases.go
```diff
@@ -96,58 +93,116 @@ func listReleasesPreconditions(args *listReleasesArguments) error {
...
-		for _, release := range result.releases {
+	// table headers
+	output := []string{strings.Join([]string{
+		color.CyanString("VERSION"),
+		color.CyanString("STATUS"),
+		color.CyanString("KUBERNETES"),
+		color.CyanString("CONTAINERLINUX"),
+		color.CyanString("COREDNS"),
+		color.CyanString("CALICO"),
+		color.CyanString("CREATED"),
```
marians: I would move `CREATED` to the third position

commands/list_releases.go
```diff
@@ -96,58 +93,116 @@ func listReleasesPreconditions(args *listReleasesArguments) error {
...
 
+		if status == "active" {
+			output = append(output, strings.Join([]string{
+				color.YellowString(*release.Version),
+				color.YellowString(status),
+				color.YellowString(kubernetes_version),
+				color.YellowString(container_linux_version),
+				color.YellowString(coredns_version),
+				color.YellowString(calico_version),
+				color.YellowString(created),
```
marians: see above regarding position of `created`

marians:
Looks very nice! ❤️ 

I haven't tested yet, but plan to do so tomorrow.

One thing: If you think Masterminds/semer can replace coreos/go-semver/semver, can we then replace it everywhere and remove coreos/go-semver/semver from the dependencies completely? Otherwise I'd ask why we need both.

tuommaki (approved):
In general this LGTM. Couple nitpicks:

* In testcases there was slight misalignation in test data.
* In general we have been strict on comment style:
  * Full sentences, starting with capital letter unless it's a Godoc for private name.
  * Period in the end of the sentence.

Error handling didn't adhere our use of `microerror`, but I guess this is due to old codebase that hasn't been refactored yet? I'm ok with this when it's consistent with other code.

commands/list_releases.go
teemow: I had this in the first place. But that isn't easy to parse and imo it is "learned" that such dates are at the end of the line. But thats your call. You still prefer it?

screen_2018-10-04-09-37-46

teemow: Maybe if I remove the time? Not sure if the time is really valuable information and we still display the exact time in `show release`

commands/show_release.go
teemow: This was in fact copied already from the `show cluster` command. I didn't do anything here.

See https://github.com/giantswarm/gsctl/blob/master/commands/show_cluster.go#L161

teemow:
> One thing: If you think Masterminds/semer can replace coreos/go-semver/semver, can we then replace it everywhere and remove coreos/go-semver/semver from the dependencies completely? Otherwise I'd ask why we need both.

I am fine with replacing it everywhere. But didn't feel responsible to do so in this PR. Let me try if its a drop-in replacement.

teemow:
> Let me try if its a drop-in replacement.

No it is not. I'll leave it as is.

teemow:
@calvix I agree but we currently don't have the provider information in here. This has been changed in the Release CRD already. So we would need to expose this via the API.

See: https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/core/v1alpha1/release_types.go#L105

The only other option would be to hard code all current providers and then try to find the operators in the components list. Shall we do this as a temporary workaround? Or shall we rather use the time to expose it properly?

marians:
```
$ ./gsctl list releases -e geckon
VERSION  STATUS      CREATED                 KUBERNETES  CONTAINERLINUX  COREDNS  CALICO
0.3.0    -           2017 Oct 26, 12:00 UTC  1.8.1       n/a             n/a      2.6.2
1.2.0    deprecated  2017 Dec 12, 12:00 UTC  1.8.4       n/a             n/a      2.6.2
1.4.0    deprecated  2018 Feb 08, 12:00 UTC  1.8.4       1576.5.0        n/a      2.6.2
2.7.0    deprecated  2018 Apr 25, 12:00 UTC  1.10.1      1688.5.3        1.1.1    3.0.5
2.9.0    deprecated  2018 Jun 01, 13:00 UTC  1.10.1      1688.5.3        1.1.1    3.0.5
2.10.0   deprecated  2018 Jun 01, 15:00 UTC  1.10.4      1688.5.3        1.1.1    3.0.5
2.11.0   deprecated  2018 Aug 09, 12:00 UTC  1.10.4      1688.5.3        1.1.1    3.0.5
2.12.0   deprecated  2018 Aug 29, 15:00 UTC  1.10.4      1688.5.3        1.1.1    3.0.5
2.12.1   active      2018 Sep 17, 14:00 UTC  1.10.4      1688.5.3        1.1.1    3.0.5
3.0.0    deprecated  2018 Aug 09, 13:00 UTC  1.11.1      1688.5.3        1.1.1    3.0.5
```

The STATUS column shows `-` for version 0.3.0. I think it should show `deprecated` there.

calvix:
@teemow Oh, that sucks. If its easy to parse the info then we can do it but probably not invest too much time into that

commands/show_release.go
teemow: fixed and also made the ReleaseVersionMissing error nicer. :+1:

teemow:
@marians i've fixed your comments. shall we keep the full date in the third column? fishing for a :+1: ?

marians (approved):
LGTM this way

PR #301, opened 2018-10-05
some minor changes to the release process

teemow:
* added the user to all build containers so the uid and gid are the same
as the user on the host
* changed the `gsctl-release` profile to `giantswarm` as it is imo more
common in the company
* switched from https repos to ssh in the package manager scripts


pipo02mix (approved):
LGTM

MarcelMue (approved):
LGTM

marians:
Thanks @teemow! Why the change in the git commands from `https://` to `git:`?

teemow:
@marians `git:` works via SSH and hence uses your SSH key. Which is most likely unlocked in your agent. `https://` uses username and password. I don't have my password in plaintext in my git config. I don't even know my password. So imo SSH is easier to use.

PR #302, opened 2018-10-08
Wait if the token was issued in the 'future'

oponder:
When gsctl detects that the tokens `iat` claim is in the future, it will wait until the computer catches up to that time.

```
$ go build && ./gsctl list clusters

Your computer's clock appears to be a bit behind.
Issued at:  1538995180
Now:        1538995165
Difference: 15 seconds

Waiting...

ID     ORGANIZATION   NAME             RELEASE  CREATED
...
```

Towards: https://github.com/giantswarm/giantswarm/issues/4314

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -398,6 +398,7 @@ func (c *configStruct) AuthHeaderGetter(endpoint string, overridingToken string)
 // whether or not the access token is valid. Does not check if the signature is valid.
 // Only checkes time based claims.
 func isTokenValid(token string) (expired bool) {
+	return false
 	// Parse token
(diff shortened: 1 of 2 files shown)


oponder:
> How about limiting the maximum wait time to e. g. 5 seconds?

I thought about limiting it, but the user could decide on their own right? They can ctrl-c at any time. 

When it's longer than 5 seconds and we kill the command, is that a better UX? Why?

I think in any case we could give some information about what they _could_ do to avoid seeing it in the future, like "Please make sure your clock is synchronized" ?

marians (approved):
Fine with it as it is for now. Let's test.

PR #304, opened 2018-10-08
Update dependencies

marians:
This updates all dependencies and fixes the `clienterror` package to update detection of a timeout error. Seems like the underlying client now reports HTTP timeout errors as type `url.Error`.


pipo02mix (approved):
LGTMIIWFY

PR #309, opened 2018-10-26
Use golang 1.11

marians:
This sets the Go version in builds to 1.11.1.

In addition, this sets a `/.cache` volume for the build, to prevent this error:

```
go: disabling cache (/.cache/go-build) due to initialization failure: mkdir /.cache: permission denied
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,43 +36,59 @@ crosscompile: build/bin/$(BIN)-darwin-amd64 build/bin/$(BIN)-linux-amd64 build/b
 # platform-specific build
 build/bin/$(BIN)-darwin-amd64: $(SOURCE)
 	@mkdir -p build/bin
-	docker run --rm -v $(shell pwd):/go/src/github.com/$(ORGANISATION)/$(PROJECT) \
+	@mkdir -p go-build-cache
+	docker run --rm \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -36,43 +36,59 @@ crosscompile: build/bin/$(BIN)-darwin-amd64 build/bin/$(BIN)-linux-amd64 build/b
 # platform-specific build
 build/bin/$(BIN)-darwin-amd64: $(SOURCE)
 	@mkdir -p build/bin
-	docker run --rm -v $(shell pwd):/go/src/github.com/$(ORGANISATION)/$(PROJECT) \
+	@mkdir -p go-build-cache
+	docker run --rm \
+		-v $(shell pwd):/go/src/github.com/$(ORGANISATION)/$(PROJECT) \
+		-v $(shell pwd)/go-build-cache:/.cache \
```
JosephSalisbury: I feel like it makes more sense to mount the local golang build cache

marians: That would only work for the architecture/os the developer is currently, on, but we are building for several architectures/os.

JosephSalisbury: I am fairly certain that the cache is smart enough to not confuse arches

PR #316, opened 2018-12-06
Automate release

marians:
Towards https://github.com/giantswarm/giantswarm/issues/4810


marians:
Now that this is merged, I'll push a new tag and see if we actually get a release, as expected.

PR #321, opened 2018-12-10
Fix a warning in the Makefile

marians:
Aftermath of https://github.com/giantswarm/giantswarm/issues/4810

This fixes a warning about the non-existing `VERSION` file (which is created in CI)


pipo02mix (approved):
LGTM

PR #324, opened 2019-01-10
Add method to fetch cluster status from new endpoint

marians:
Towards https://github.com/giantswarm/giantswarm/issues/5037

This adds a function to fetch cluster status information from the new [cluster status endpoint](https://docs.giantswarm.io/api/#operation/getClusterStatus).

Adds an additional HTTP client `rawClient` to the WrapperV2, apart from the API client generated based on OpenAPI spec, so we can parse the response in arbitrary ways.

`gsctl show cluster` makes use of that function in order to determine the current number of nodes.

## List of changes

- Adds method to fetch cluster status from new status endpoint
- Make use of worker node count from new cluster status info in `show cluster` command
- `gsctl show cluster` output:
  - Show line `Worker nodes scaling` with either `pinned to <n>` or `autoscaling between <min> and <max>` to inform about the current worker node scaling limits.
  - Display the current actual number of nodes, as per the new status endpoint, as `Worker nodes running`.
  - If `scaling.min` and `scalling.max` differ, show `Desired worker node count` row, which indicates the number of worker nodes the autoscaler currently tries to achieve.
- `gsctl info` output:
  - Aligned labelling of rows with `show cluster` output. `Worker instance type options` changes to `Worker EC2 instance type options`. `Default worker instance type` changes to `Default worker EC2 instance type`.


pipo02mix (changes requested):
Change how workers count is get

marians:
Found a problem to fix

image

MarcelMue:
@marians I think tuomas mentioned that this is still an issue, uncertain if he is actively fixing this.

marians:
The screenshot above just illustrates that the mechanism with subtracting one node (master) from the slice length can lead to a negative worker count, as the status endpoint doesn't show nodes right after cluster creation.

client/client.go
```diff
@@ -72,6 +75,23 @@ type WrapperV2 struct {
...
+
+// ClusterStatus is a type we use to unmarshal a cluster status JSON response
+// from the API. Note: this is scarce, leaving out many available details.
+type ClusterStatus struct {
+	Cluster *ClusterStatusCluster `json:"cluster,omitempty"`
+}
+type ClusterStatusCluster struct {
+	Nodes []*ClusterStatusClusterNode `json:"nodes,omitempty"`
+}
+type ClusterStatusClusterNode struct {
```
xh3b4sd: Why don't we import `apiextensions` here and use these types? This is what defines them.

client/client.go
```diff
@@ -371,6 +397,42 @@ func (w *WrapperV2) GetCluster(clusterID string, p *AuxiliaryParams) (*clusters.
 	return response, nil
 }
 
+// GetClusterStatus fetches status details from the
+// status endpoint (/v4/clusters/{clusterid}/status/).
+// Only a few details are returned.
+func (w *WrapperV2) GetClusterStatus(clusterID string) (*ClusterStatus, *http.Response, error) {
+	url := w.conf.Endpoint + fmt.Sprintf("/v4/clusters/%s/status/", clusterID)
```
xh3b4sd:
```suggestion
	url := fmt.Sprintf("%s/v4/clusters/%s/status/", w.conf.Endpoint, clusterID)
```

commands/show_cluster.go
```diff
@@ -181,35 +209,52 @@ func showClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
...
-	if err == nil && clusterDetails.CredentialID != "" {
-		if args.verbose {
-			fmt.Println(color.WhiteString("Fetching details for credential %s", clusterDetails.CredentialID))
-		}
-
-		credentialDetails, err = getOrgCredentials(clusterDetails.Owner, clusterDetails.CredentialID, showClusterActivityName)
+	clusterDetailsChan := make(chan *models.V4ClusterDetailsResponse)
+	clusterDetailsErrChan := make(chan error)
+
+	go func(chan *models.V4ClusterDetailsResponse, chan error) {
```
xh3b4sd: What is all this supposed to do? When your goal is to run different actions in parallel, please use https://godoc.org/golang.org/x/sync/errgroup.

client/client.go
marians: Done

commands/show_cluster.go
```diff
@@ -220,6 +265,23 @@ func showClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
...
+
+		credentialDetails, _ = getOrgCredentials(clusterDetails.Owner, clusterDetails.CredentialID, showClusterActivityName)
+	}
+
+	// Calculate worker count: if status info contains Cluster.Nodes, we use that.
+	// Otherwise fall back to old style workers slice.
+	numWorkers := len(clusterDetails.Workers)
+	if clusterStatus != nil && clusterStatus.Cluster != nil && clusterStatus.Cluster.Nodes != nil {
+		numWorkers = len(clusterStatus.Cluster.Nodes) - 1 // assuming one master node to subtract
+	}
```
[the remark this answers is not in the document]
marians: Labels are not yet available in the API. Added a `TODO` marker. If this is available early enough, I'll change.

commands/show_cluster.go
marians: Looked at it, but found that `errgroup's` `Group.Wait` would only return the first error happening. That doesn't help me in this case here, as I want to treat an error happening when fetching cluster details (which is the critical request) different from errors when fetching cluster status.

marians:
Can this get an approval? @pipo02mix

pipo02mix (dismissed):
I am ok but not sure about the error handling

tuommaki (approved):
I think this is mostly ok. I think the concurrent cluster detail & status fetching could be improved for readability in general, but I cannot suggest anything quick in here so I guess that's something for later iteration and general overhaul of gsctl style wise.

client/client.go
```diff
@@ -371,6 +391,42 @@ func (w *WrapperV2) GetCluster(clusterID string, p *AuxiliaryParams) (*clusters.
 	return response, nil
 }
 
+// GetClusterStatus fetches status details from the
+// status endpoint (/v4/clusters/{clusterid}/status/).
+// Only a few details are returned.
+func (w *WrapperV2) GetClusterStatus(clusterID string) (*ClusterStatus, *http.Response, error) {
```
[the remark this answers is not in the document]
tuommaki: I'd maybe prefer this rather than creating a separate client. Maybe add proper explanation for the Marshal-Unmarshal-dance in a comment so that someone who might wonder about this gets the story as well.

pipo02mix (approved):
I am happy too

client/client.go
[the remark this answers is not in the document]
marians: Looks good to me. Yes, commenting for the future colleague trying to wrap her/his hand around this would be nice.

PR #325, opened 2019-01-14
Support cluster autoscaling

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2206. This PR is the final feature branch in which we merge multiple separate changes. At the end of the feature implementation we merge this PR in order to be done.

## List of changes

To be used for the gsctl release notes

---

This release of gsctl focuses on support new API features introduced to enable cluster autoscaling on AWS for releases starting at 6.3.0.

### Major changes

- `gsctl create cluster`
  - Two new (optional) flags `--workers-min` and `--workers-max` are introduced to set the initial scaling settings for the number of worker nodes in a cluster. The autoscaler will control the number of worker nodes within this range. For AWS releases prior to 6.3.0, and on Azure and KVM, both flags must be set to the same value. As before, once a cluster is created, the worker node count (or the range) cna be influenced using `gsctl scale cluster`.
  - The flag `--num-workers` can still be used to set the initial cluster size, as an alternative to `--workers-min` and `--workers-max`. The value given will be applied to both the minimum and maximum cluster size, effectively disabling autoscaling.
- `gsctl scale cluster`
  - The new flags `--workers-min` and `--workers-max` are introduced to set the scaling limits for a cluster. On releases supporting autoscaling (AWS <= 6.3.0), they can be used independent of each other to adjust only the upper or lower limit.
  - The `--num-workers`/`-w` flag can still be used. In releases supporting autoscaling, tis will result in setting the minimum and maximum worker node count to the same value, effectively disabling autoscaling.
  - The flags `--memory-gb`, `--num-cpus`, and `--storage-gb` got removed as they were no longer functional
- `gsctl show cluster`
  - The new line `Worker node scaling` is added, informing about the current scaling settings.
  - The new line `Worker nodes running` replaces the former `Workers` line, informing about the number of worker nodes running.
  - The new line `Desired worker node count` is shown when the cluster supports autoscaling and the worker node count is not pinned to a specific number. It shows the number of worker nodes the autoscaler attempts to have running currently.


marians:
Can you please adjust the dep settings so that non-Go files are pruned? Just noticed that with the `backoff` library.

xh3b4sd:
I am not sure what you mean. This seems to be irrelevant for this branch here. Let's talk or feel free to create a PR based on `master` with the suggestions you would like to see.

marians:
I mean files like https://github.com/giantswarm/gsctl/blob/083d32dae6286844ad40dbf323a440b06b19a95c/vendor/github.com/cenkalti/backoff/.travis.yml - I just happened to notice them in this PR and thought that you could pull the `dep` config change out of your hat. But would be welcome in a different PR as well.

xh3b4sd:
AFAICT the prune settings are the same we have in all other repos. When this is a problem we probably have it everywhere and would need to decide if we want to address it at all when there are go modules not so far away on the horizon.

jgsqware:
As soon this is merged, I will add this in the Cluster Resource Usage tool: ref https://github.com/giantswarm/giantswarm/issues/5257

marians (approved):

PR #326, opened 2019-01-16
first changes for cluster creation towards auto scaling

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2206. This branch is based on [`feature-auto-scaling`](https://github.com/giantswarm/gsctl/pull/325). The specs for command UX design can be found at https://github.com/giantswarm/giantswarm/issues/5037. 

**Note that the current changes are WIP and I only pinged stakeholders for knowledge sharing.** I will ask later again for reviews. I still have to test and further implement a couple of things.

### DIFF
--- a/commands/create_cluster.go
+++ b/commands/create_cluster.go
@@ -440,8 +440,8 @@ func createDefinitionFromFlags(args addClusterArguments) clusterDefinition {
 
 	if args.numWorkers != 0 {
 		workers := []nodeDefinition{}
-		for i := 0; i < args.numWorkers; i++ {
 
+		for i := 0; i < args.numWorkers; i++ {
 			worker := nodeDefinition{}
 
 			if args.workerNumCPUs != 0 {
@@ -468,8 +468,14 @@ func createDefinitionFromFlags(args addClusterArguments) clusterDefinition {
 
 			workers = append(workers, worker)
 		}
+
 		def.Workers = workers
+		if def.Scaling.Min == 0 && def.Scaling.Max == 0 {
+			def.Scaling.Min = int64(len(def.Workers))
+			def.Scaling.Max = int64(len(def.Workers))
+		}
 	}
+
 	return def
 }
 
--- a/commands/create_cluster.go
+++ b/commands/create_cluster.go
@@ -30,6 +30,7 @@ type addClusterArguments struct {
 	inputYAMLFile           string
 	numWorkers              int
 	owner                   string
+	releaseVersion          string
 	scheme                  string
 	token                   string
 	wokerAwsEc2InstanceType string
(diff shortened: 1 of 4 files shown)


commands/create_cluster.go
```diff
@@ -53,14 +55,16 @@ func defaultAddClusterArguments() addClusterArguments {
...
+		releaseVersion:          cmdRelease,
 		scheme:                  scheme,
 		token:                   token,
-		releaseVersion:          cmdRelease,
 		wokerAwsEc2InstanceType: cmdWorkerAwsEc2InstanceType,
 		wokerAzureVMSize:        cmdWorkerAzureVMSize,
 		workerNumCPUs:           cmdWorkerNumCPUs,
 		workerMemorySizeGB:      cmdWorkerMemorySizeGB,
 		workerStorageSizeGB:     cmdWorkerStorageSizeGB,
+		workersMax:              cmdWorkersMax,
```
MarcelMue: The old var was called `numWorkers`, would it make sense to call the new ones `numWorkersMax` accordingly?

commands/create_cluster.go
```diff
@@ -30,14 +30,16 @@ type addClusterArguments struct {
...
+	releaseVersion          string
 	scheme                  string
 	token                   string
 	wokerAwsEc2InstanceType string
 	wokerAzureVMSize        string
 	workerNumCPUs           int
 	workerMemorySizeGB      float32
 	workerStorageSizeGB     float32
-	releaseVersion          string
+	workersMax              int64
```
MarcelMue: Why int64 instead of just `int` ? I assume this has to do with the generated client?

commands/create_cluster.go
```diff
@@ -178,6 +189,12 @@ func createClusterValidationOutput(cmd *cobra.Command, args []string) {
 		case IsConflictingFlagsError(err):
 			headline = "Conflicting flags used"
 			subtext = "When specifying a definition via a YAML file, certain flags must not be used."
+		case IsConflictingWorkerFlagsUsed(err):
```
MarcelMue: Kink of unrelated: I will need to use this error as well - is there a way to share headline / subtext right now? Otherwise I'll just have to copy it to scaling I guess.

commands/create_cluster.go
xh3b4sd: The generated client works with `int64` and I preferred to work with one data type instead of doing conversions somewhere.

commands/create_cluster.go
xh3b4sd: Copy it and we sort the merging later.

commands/create_cluster.go
xh3b4sd: I would keep it like that for now. There are more important things to get right. My priority here is only to get the feature out of the door, as we have to put more love into this code base anyway.

MarcelMue (approved):
Let's get this merged, there is some code share between commands I'd like to use.

PR #327, opened 2019-01-16
Change scaling to min and max

MarcelMue:


MarcelMue:
I tried to test most of this manually, but the API is not quite there yet for scaling.

commands/scale_cluster.go
```diff
@@ -38,108 +38,277 @@ Examples:
...
+		fmt.Println(color.RedString("Error getting cluster status!"))
+		handleCommonErrors(err)
+		fmt.Println(color.RedString(err.Error()))
+		os.Exit(1)
+	}
+
+	var maxBefore int64
+	var minBefore int64
+	var desiredCapacity int64
+	// Older clusters without scaling shouldn't have scaling values but have a workers array instead.
```
MarcelMue: I decided against checking the cluster version here and instead checking if there are scaling settings present. Is this okay to do or should I do the semver comparison instead here?

commands/scale_cluster.go
```diff
@@ -38,108 +38,277 @@ Examples:
...
 		Run: scaleClusterRunOutput,
 	}
+
+	//Flag names.
+	cmdWorkersMaxName            = "workers-max"
+	cmdWorkersMinName            = "workers-min"
+
+	cmdWorkerMemorySizeGBName    = "memory-gb"
+	cmdWorkerNumCPUsName         = "num-cpus"
+	cmdWorkerStorageSizeGBName    = "storage-gb"
```
xh3b4sd: That looks like `gofmt` should be run.

commands/scale_cluster.go
```diff
@@ -38,108 +38,277 @@ Examples:
...
-	numWorkersToAdd int
-	// number of workers as of the PATCH call response.
-	numWorkersAfter int
+	if !cmd.Flags().Changed(cmdWorkersMinName) {
+		scaleArgs.workersMin = minBefore
+	}
+	if !cmd.Flags().Changed(cmdWorkersMaxName) {
+		scaleArgs.workersMax = maxBefore
+	}
+	if !cmd.Flags().Changed(cmdWorkersMaxName) && !cmd.Flags().Changed(cmdWorkersMinName) && cmd.Flags().Changed("num-workers") {
```
xh3b4sd: What about having a constant for `num-workers` as well?

commands/scale_cluster.go
```diff
@@ -38,108 +38,277 @@ Examples:
...
+func getClusterStatus(clusterID, activityName string) (*v1alpha1.StatusCluster, error) {
+	// perform API call
+	auxParams := ClientV2.DefaultAuxiliaryParams()
+	auxParams.ActivityName = activityName
 
-	ScaleCommand.AddCommand(ScaleClusterCommand)
-}
+	response, err := ClientV2.GetClusterStatus(clusterID, auxParams)
+	if err != nil {
+		if clientErr, ok := err.(*clienterror.APIError); ok {
```
xh3b4sd: I would prefer having the usual error matchers as defined in https://github.com/giantswarm/fmt/blob/master/go/errors.md.

commands/scale_cluster.go
xh3b4sd: I think the version checks would be the more proper way of doing it. Same for what we do in the API.

commands/scale_cluster_test.go
```diff
@@ -33,7 +33,7 @@ func TestScaleClusterNotLoggedIn(t *testing.T) {
 	cmdAPIEndpoint = mockServer.URL
 	initClient()
 
-	err := verifyScaleClusterPreconditions(testArgs, []string{testArgs.clusterID})
+	err := validateScaleCluster(testArgs, []string{testArgs.clusterID}, 5, 5, 5)
```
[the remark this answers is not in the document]
xh3b4sd: It would be pretty cool to have an isolated test similar to what I try to do in the API.

tuommaki (approved):

tuommaki:
Right. GitHub hid the major diff out of sight again. Oh well, Tim got you covered so check his comments. Otherwise I guess this looked good. I already approved it ¯\_(ツ)_/¯

client/client.go
```diff
@@ -482,3 +482,19 @@ func (w *WrapperV2) SetCredentials(organizationID string, addCredentialsRequest
...
+		return nil, microerror.Mask(err)
+	}
+
+	response, err := w.gsclient.Clusters.GetClusterStatus(params, nil)
+	if err != nil {
+		return nil, clienterror.New(err)
+	}
+
+	return response, nil
+}
```
marians: Wouldn't it be nice if this function could do the unmarshal/marshal dance and return the usable struct instead of the not so usable map[string]interface?

marians: In other words, should we move https://github.com/giantswarm/gsctl/blob/scaling-update/commands/scale_cluster.go#L175-L188 over here?

commands/scale_cluster_test.go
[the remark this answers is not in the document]
MarcelMue: I'd like to add tests in that style in another PR. This is already pretty big and there will have to be some structure changes to make it work similar to the API.

commands/scale_cluster.go
MarcelMue: I added a version check like in the API.

commands/scale_cluster.go
MarcelMue: Told my editor to behave.

commands/scale_cluster.go
MarcelMue: Done.

commands/scale_cluster.go
MarcelMue: I used the fmt way error matching now.

client/client.go
MarcelMue: I tried this, but then we get a weird situation:
Everything in `client.go` returns the swagger generated types, but now one function would return the apiextensions type. It would then also no longer be `response.Payload` but just `response`. It seems equally inconsistent than just doing it in this function.

I'll move it if it's bothering you, but I don't see a big benefit.

commands/scale_cluster.go
```diff
@@ -38,115 +39,307 @@ Examples:
...
 
-	ScaleCommand.AddCommand(ScaleClusterCommand)
-}
+	response, err := ClientV2.GetClusterStatus(clusterID, auxParams)
+	if err != nil {
+		if IsAPIError(err) {
+			clientErr, ok := err.(*clienterror.APIError)
+			if ok {
+				switch clientErr.HTTPStatusCode {
+				case http.StatusForbidden:
```
xh3b4sd: I would actually use matchers for all these cases as well to just have the flat error checks hierarchy as suggested below. Without all the conditional indentions. We could have 1 instead of 4 indentation levels. 

```
if IsStatusForbiddenError(err) {
  ...
} else if IsStatusUnauthorized(err) {
  ...
} else if IsStatusNotFound(err) {
  ...
} else if IsStatusInternalServerError(err) {
  ...
} else if err != nil {
  ...
}
```

commands/scale_cluster.go
```diff
@@ -38,115 +39,307 @@ Examples:
...
+	if len(cmdLineArgs) == 0 {
+		handleCommonErrors(clusterIDMissingError)
+	}
+	// Get all necessary information for defaulting and confirmation.
+	clusterDetails, err := getClusterDetails(cmdLineArgs[0], scaleClusterActivityName)
+	if err != nil {
+		fmt.Println(color.RedString("Error getting cluster details!"))
+		handleCommonErrors(err)
+		fmt.Println(color.RedString(err.Error()))
+		os.Exit(1)
```
xh3b4sd: We should start wrapping functionality and have proper error handling. The caller should then be the only place where the program is killed. Having this all over the place is super messy. We should aim for one line error handling. Not four. We would then also improve maintainability and testability.

PR #329, opened 2019-01-23
Only send one worker on cluster creation

MarcelMue:
We discussed this yesterday:
The behaviour is supposed to be, that we only send one worker when creating a cluster - no matter which version it is in.

The API does not currently support this, have we changed our thinking here or is this still valid?

### DIFF
--- a/commands/create_cluster.go
+++ b/commands/create_cluster.go
@@ -470,15 +470,17 @@ func createAddClusterBody(d clusterDefinition) *models.V4AddClusterRequest {
 		Max: d.Scaling.Max,
 	}
 
-	for _, dWorker := range d.Workers {
+	if len(d.Workers) >= 1 {
 		ndmWorker := &models.V4AddClusterRequestWorkersItems{}
-		ndmWorker.Memory = &models.V4AddClusterRequestWorkersItemsMemory{SizeGb: float64(dWorker.Memory.SizeGB)}
-		ndmWorker.CPU = &models.V4AddClusterRequestWorkersItemsCPU{Cores: int64(dWorker.CPU.Cores)}
-		ndmWorker.Storage = &models.V4AddClusterRequestWorkersItemsStorage{SizeGb: float64(dWorker.Storage.SizeGB)}
-		ndmWorker.Labels = dWorker.Labels
-		ndmWorker.Aws = &models.V4AddClusterRequestWorkersItemsAws{InstanceType: dWorker.AWS.InstanceType}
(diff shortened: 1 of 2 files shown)


commands/create_cluster.go
```diff
@@ -470,15 +470,17 @@ func createAddClusterBody(d clusterDefinition) *models.V4AddClusterRequest {
 		Max: d.Scaling.Max,
 	}
 
-	for _, dWorker := range d.Workers {
+	if len(d.Workers) >= 1 {
```
xh3b4sd: In which cases is it greater than 1?

MarcelMue: There is still some old behaviour that creates more than one worker with the `--num-workers` flag. I'll sort it out. Thanks for reading carefully!

commands/create_cluster.go
```diff
@@ -417,43 +415,34 @@ func definitionFromFlags(def clusterDefinition, args addClusterArguments) cluste
...
-				worker.Azure.VMSize = args.wokerAzureVMSize
-			}
+	if def.Scaling.Min == 0 && def.Scaling.Max == 0 {
+		def.Scaling.Min = int64(len(def.Workers))
+		def.Scaling.Max = int64(len(def.Workers))
+	}
 
-			workers = append(workers, worker)
-		}
+	workers := []nodeDefinition{}
```
MarcelMue: We always only create one worker.

commands/create_cluster.go
```diff
@@ -536,29 +527,14 @@ func addCluster(args addClusterArguments) (addClusterResult, error) {
...
 		fmt.Printf("Requesting new cluster for organization '%s'\n", color.CyanString(result.definition.Owner))
-
 		auxParams := ClientV2.DefaultAuxiliaryParams()
 		auxParams.ActivityName = createClusterActivityName
-
 		// perform API call
 		response, err := ClientV2.CreateCluster(addClusterBody, auxParams)
 		if err != nil {
-			// create specific error types for cases we care about
-			if clientErr, ok := err.(*clienterror.APIError); ok {
```
MarcelMue: Okay, I took my time and had a closer look at what is happening:
I assumed that `APIError` is a generated type, but it is actually an error type we define here: https://github.com/giantswarm/gsctl/blob/master/client/clienterror/clienterror.go
So the thing we effectively do here is compare an error type (`APIError`) which we create to another microerror error type (from https://github.com/giantswarm/gsctl/blob/master/commands/error.go) which we also create. 

The refactoring goal would be to move the commands to different packages and then only use the `microerror` type with proper matchers instead of defining everything twice.

PR #330, opened 2019-01-23
Check for individual errors explicitly

MarcelMue:
Now we can have much more flat error checking.

### DIFF
--- a/commands/error.go
+++ b/commands/error.go
@@ -1,6 +1,8 @@
 package commands
 
 import (
+	"net/http"
+
 	"github.com/giantswarm/gsctl/client/clienterror"
 	"github.com/giantswarm/microerror"
 )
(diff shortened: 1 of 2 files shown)


MarcelMue:
I talked to @marians - the previous implementation didn't do anything because it is superseded by the `handleCommonErrors` function. So if anything it gave us a worse UX output of the error.

I removed the use of the matchers, but would like to keep the improved matchers.

MarcelMue:
We should go through the code when we get to do some more widespread refactoring, there are more instances where we unnecessarily rewrite the errors.

xh3b4sd:
Well, this common handling is a bit off and makes the code worse IMO. The version you had was good. The specific error types could just have been used to come up with the appropriate UX feedback in the terminal.

MarcelMue:
Yes, but we need to fix the common handling to do anything. The version I had was effectively not doing anything. I want to refactor the common error handling to be more in line with what we usually do, just not in this PR.

The common handling prints the `Message` + `Details` which the types don't have right now. There are also many more cases then the few that were here. The refactoring goal is going to be to improve on this (not having this common thing anymore), BUT having dead code all over the place makes it really hard to see what is actually happening right now though.

So my plan: 
1. Get rid of dead code
2. Improve and refactor error handling

PR #331, opened 2019-01-24
Get the status only when we are sure that it holds useful information

MarcelMue:
This is really just about saving us an unnecessary api call.

### DIFF
--- a/commands/scale_cluster.go
+++ b/commands/scale_cluster.go
@@ -226,8 +226,6 @@ func scaleClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
 		}
 		if n == 0 || n == 1 {
 			autoScalingEnabled = true
-		} else {
-			autoScalingEnabled = false
 		}
 	}
 
--- a/commands/scale_cluster.go
+++ b/commands/scale_cluster.go
@@ -155,7 +155,7 @@ func getClusterStatus(clusterID, activityName string) (*v1alpha1.StatusCluster, 
 	auxParams.ActivityName = activityName
 
 	response, err := ClientV2.GetClusterStatus(clusterID, auxParams)
- 	if err != nil{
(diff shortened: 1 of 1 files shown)


commands/scale_cluster.go
```diff
@@ -196,8 +196,8 @@ func scaleCluster(args scaleClusterArguments) error {
 	auxParams.ActivityName = scaleClusterActivityName
 
 	_, err := ClientV2.ModifyCluster(args.clusterID, reqBody, auxParams)
-	if err != nil{
-		return  microerror.Mask(err)
+	if err != nil {
```
xh3b4sd: Looks like somebody should get their golang editor integrations sorted. 😛

commands/scale_cluster.go
```diff
@@ -243,10 +236,23 @@ func scaleClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
 	var desiredCapacity int64
```
xh3b4sd: I cannot comment on the line I spotted above. 

```
		if n == 0 || n == 1 {
			autoScalingEnabled = true
		} else {
			autoScalingEnabled = false
		} 
```

There is no reason for the `else` clause since the zero value is already `false`.

MarcelMue: Fixed.

commands/scale_cluster.go
MarcelMue: Tell me about it, Atom is really grinding my gears the last few days.

xh3b4sd: I use https://atom.io/packages/go-plus and have `goimports` for the formatting and stuff.

MarcelMue: I have exactly the same setup, I really like `go-plus` but for some reason parts of it have randomly stopped working.

PR #333, opened 2019-01-25
Use api scaling response for user feedback on new scaling settings

MarcelMue:
We just used the values that we defaulted from the user before to show the change. Now this should be correct and actually use the API response.


xh3b4sd (approved):

pipo02mix (approved):
nice LGTM

tuommaki (approved):
:+1:

PR #336, opened 2019-01-28
Add default for missing workers 

MarcelMue:
Based on #325

### DIFF
--- a/client/clienterror/clienterror.go
+++ b/client/clienterror/clienterror.go
@@ -131,6 +131,11 @@ func New(err error) *APIError {
 			ae.ErrorDetails += "Details: " + modifyClusterFailedErr.Payload.Message
 		}
 
+		if modifyClusterFailedErr.Payload.Code == "INVALID_INPUT" {
+			ae.ErrorMessage = "Invalid parameters"
+			ae.ErrorDetails = "The cluster could not be updated. workers-min and workers-max must be equal without autoscaling."
+		}
+
 		return ae
 	}
 	if modifyClusterFailedErr, ok := err.(*clusters.ModifyClusterNotFound); ok {
--- a/commands/create_cluster.go
+++ b/commands/create_cluster.go
@@ -416,8 +416,13 @@ func definitionFromFlags(def clusterDefinition, args addClusterArguments) cluste
 	}
 
 	if def.Scaling.Min == 0 && def.Scaling.Max == 0 {
-		def.Scaling.Min = int64(len(def.Workers))
-		def.Scaling.Max = int64(len(def.Workers))
+		def.Scaling.Min = int64(args.numWorkers)
+		def.Scaling.Max = int64(args.numWorkers)
+	}
+
+	if def.Scaling.Min == 0 && def.Scaling.Max == 0 && args.numWorkers == 0 {
+		def.Scaling.Min = 3
+		def.Scaling.Max = 3
 	}
 
 	workers := []nodeDefinition{}
@@ -468,8 +473,6 @@ func createAddClusterBody(d clusterDefinition) *models.V4AddClusterRequest {
 		ndmWorker.Aws = &models.V4AddClusterRequestWorkersItemsAws{InstanceType: d.Workers[0].AWS.InstanceType}
 		ndmWorker.Azure = &models.V4AddClusterRequestWorkersItemsAzure{VMSize: d.Workers[0].Azure.VMSize}
 		a.Workers = append(a.Workers, ndmWorker)
-	} else {
-		a.Workers = append(a.Workers, &models.V4AddClusterRequestWorkersItems{})
 	}
 
 	return a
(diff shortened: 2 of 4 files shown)


client/clienterror/clienterror.go
```diff
@@ -131,6 +131,11 @@ func New(err error) *APIError {
 			ae.ErrorDetails += "Details: " + modifyClusterFailedErr.Payload.Message
 		}
 
+		if modifyClusterFailedErr.Payload.Code == "INVALID_INPUT" {
```
MarcelMue: This problem was noted here: https://github.com/giantswarm/giantswarm/pull/2206#issuecomment-457584528

Output now looks like this:
```
 gsctl scale cluster 5nkxf --workers-max 4
The cluster size is currently pinned to 3.
Do you want to change the limits to be min=3, max=4? [y/n]: y
Invalid parameters
The cluster could not be updated. workers-min and workers-max must be equal without autoscaling.
```

commands/create_cluster.go
```diff
@@ -416,8 +416,13 @@ func definitionFromFlags(def clusterDefinition, args addClusterArguments) cluste
 	}
 
 	if def.Scaling.Min == 0 && def.Scaling.Max == 0 {
-		def.Scaling.Min = int64(len(def.Workers))
-		def.Scaling.Max = int64(len(def.Workers))
+		def.Scaling.Min = int64(args.numWorkers)
```
MarcelMue: `def.Workers` is not set here yet, this solves a problem where scaling values were always `0` if only `num-workers`  was given.

commands/create_cluster.go
```diff
@@ -416,8 +416,13 @@ func definitionFromFlags(def clusterDefinition, args addClusterArguments) cluste
...
+		def.Scaling.Max = int64(args.numWorkers)
+	}
+
+	if def.Scaling.Min == 0 && def.Scaling.Max == 0 && args.numWorkers == 0 {
```
MarcelMue: We need to take care of this case here, we need to default these settings because we need to send atleast 1 worker for e.g. `cpu` and so on. The API will only create 1 worker otherwise (which is correct behaviour).

commands/create_cluster.go
```diff
@@ -507,24 +510,24 @@ func addCluster(args addClusterArguments) (addClusterResult, error) {
...
+	_, marshalErr := json.Marshal(addClusterBody)
+	if marshalErr != nil {
+		return result, microerror.Maskf(couldNotCreateJSONRequestBodyError, marshalErr.Error())
+	}
+
 	// Preview in YAML format
 	if args.verbose {
 		fmt.Println("\nDefinition for the requested cluster:")
-		d, marshalErr := yaml.Marshal(result.definition)
+		d, marshalErr := yaml.Marshal(addClusterBody)
```
MarcelMue: Just moving some things to get more complete output here.

commands/scale_cluster.go
```diff
@@ -329,7 +336,7 @@ func validateScaleCluster(args scaleClusterArguments, cmdLineArgs []string, maxB
 	}
 
 	// flag conflicts.
-	if args.numWorkersDesired > 0 && (args.workersMax > 0 || args.workersMin > 0) {
+	if args.numWorkersDesired > 0 && (args.workersMax > 0 || args.workersMin > 0) && (int64(args.numWorkersDesired) != args.workersMax || int64(args.numWorkersDesired) != args.workersMin) {
```
MarcelMue: Flags are conflicting **only** if the `numWorkers` and `workersMin` / `workersMax` is not equal and not zero.

MarcelMue (commented):
This PR should sort out all the problems noted here https://github.com/giantswarm/giantswarm/pull/2206#issuecomment-457584528

I still need to test if there were any regressions.

commands/scale_cluster.go
```diff
@@ -201,16 +214,10 @@ func scaleClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
...
-			fmt.Println(color.RedString(err.Error()))
-			os.Exit(1)
-		}
-		if n == 0 || n == 1 {
-			autoScalingEnabled = true
-		}
+	autoScalingEnabled, err := isAutoscalingEnabled(clusterDetails.ReleaseVersion)
+	if err != nil {
+		fmt.Println(color.RedString(err.Error()))
+		os.Exit(1)
```
xh3b4sd: We should stop doing this. The program should only be terminated where it got basically initiated so that start and end are the same at best. Down the road errors should be handled so every caller can decide what and how to do things. This has to improve in the future.

xh3b4sd (approved):
Go as you see fit.

commands/create_cluster.go
pipo02mix: if the default correct behavior is creating 1 worker why we use a different one?

MarcelMue: The case is the following:
- We start an old cluster without autoscaling and submit the worker array with 1 worker (to have cpu).
- API receives a request without scaling values but one worker in workers array
- API can't decide if an old gsctl wants to create a cluster with one worker or if it should default to three workers

-> solution: we send the scalign values defaulted here :)

commands/scale_cluster.go
MarcelMue: I agree, this func is currently called by cobra. There should be a `Runner` similar to how Pawel handled it in devctl. Changing this right now is kind of a refactoring step I want to do later.

pipo02mix (approved):
LGTM beyond Tim's comment about conditional clarity

commands/scale_cluster.go
MarcelMue: Would it improve it to have the conditions in multiple lines? I can't really think of an improvement without changing a lot of the structure (which I would rather not do as some of it is shared with other commands).

xh3b4sd: We want to reduce complexity to make it understandable. I am interested in when this condition holds true. So, which case does it cover? We can have speaking methods like `args.isScalingEmpty`. That would reduce complexity for `(args.workersMax > 0 || args.workersMin > 0)`, which is already a great deal. 

Then we can use https://github.com/giantswarm/ruleengine or write something like this. 

```
isTrue := true && true
isFalse := false && false

if isTrue && isFalse {
    ...
}
```

tuommaki:
```suggestion
    desiredWorkersExists := (args.numWorkersDesired > 0)
    scalingParameterIsPresent := (args.workersMax > 0 || args.workersMin > 0)
    desiredWorkersDifferFromMaxNumOfWorkers := (int64(args.numWorkersDesired) != args.workersMax)
    desiredWorkersDifferFromMinNumOfWorkers := (int64(args.numWorkersDesired) != args.workersMin)
    desiredWorkersAreNotAtScalingLimits := (desiredWorkersDifferFromMaxNumOfWorkers || desiredWorkersDifferFromMinNumOfWorkers)

    if desiredWorkersExists && scalingParameterIsPresent && desiredWorkersAreNotAtScalingLimits {
```

Possibly something like this?

MarcelMue: I followed Tuomas suggestion here for the most part as I think it is a decent solution.
I tried to use the parameters in other validations where it made sense.

PR #338, opened 2019-02-04
Prevent error when cluster status endpoint cannot be fetched

marians:
Part of autoscaling changes.

This prevents the error info in the top of the `show cluster` output that occurred when the cluster status endpoint could not yet be accessed.

Here is how it will appear for a cluster after a creation period:

```nohighlight
...
Worker node scaling:       pinned at 0
Worker nodes running:      0
CPU cores in workers:      0
RAM in worker nodes (GB):  0.00

Info: Could not fetch cluster status, so the worker node count displayed might derive from the actual number.
```

Within 20 minutes from creation, an additional line will be shown:

```nohighlight
...
Worker node scaling:       pinned at 0
Worker nodes running:      0
CPU cores in workers:      0
RAM in worker nodes (GB):  0.00

Info: Could not fetch cluster status, so the worker node count displayed might derive from the actual number.
This is expected for clusters which are still in creation.
```

### DIFF
--- a/commands/show_cluster.go
+++ b/commands/show_cluster.go
@@ -342,7 +342,7 @@ func showClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
 		fmt.Println("\nInfo: Could not fetch cluster status, so the worker node count displayed might derive from the actual number.")
 
 		if time.Since(created) < clusterCreationExpectedDuration {
-			fmt.Println("This is expected for clusters which are still in creation.")
+			fmt.Println("This is expected for clusters which are most likely still in creation.")
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


commands/show_cluster.go
```diff
@@ -337,4 +335,14 @@ func showClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
...
 
 	fmt.Println(columnize.SimpleFormat(output))
+
+	// Here we inform about problems fetching the cluster status, which happens regularly for new clusters
+	// and isn't a problem.
+	if clusterStatusErr != nil {
+		fmt.Println("\nInfo: Could not fetch cluster status, so the worker node count displayed might derive from the actual number.")
+
+		if time.Since(created) < clusterCreationExpectedDuration {
+			fmt.Println("This is expected for clusters which are still in creation.")
```
MarcelMue:
```suggestion
			fmt.Println("This is expected for clusters which are most likely still in creation.")
```

PR #340, opened 2019-02-05
add proper defaulting for scaling

xh3b4sd:
```
$ gsctl show cluster 5o6z9
...
Release version:           5.2.2
Worker node scaling:       pinned at 4
...
```

```
$ gsctl scale cluster 5o6z9 --workers-min 3
The cluster size is currently pinned to 4.
Do you want to pin the number of worker nodes to 3? [y/n]: n
command aborted error
```

```
$ gsctl scale cluster 5o6z9 --workers-max 3
The cluster size is currently pinned to 4.
Do you want to pin the number of worker nodes to 3? [y/n]: n
command aborted error
```


cmd/cluster/scale/defaulting/scaling.go
```diff
@@ -0,0 +1,131 @@
...
+				func() bool { return s.desiredScalingMaxChanged },
+			},
+			Defaulting: func() {
+				scaling.Max = s.desiredScalingMax
+			},
+		},
+		{
+			Conditions: []func() bool{
+				func() bool { return s.autoScalingEnabled },
+				func() bool { return !s.desiredScalingMinChanged },
```
tuommaki: Just a note while reviewing this: Spotting difference between this block and first block was an exercise and made me initially question if there's something wrong when `scaling.Min` changes twice in a row - until I spotted `!`.

What do you think if there was a function for this kind of cases in order to improve readability? The above function could be then:
```go
func() bool { return not(s.desiredScalingMingChanged) },
```

or if we go the rabbit hole even deeper:
```go
func() bool { return when(not(s.desiredScalingMingChanged)) },
```

The helper functions could be:
```go

func not(v bool) bool {
    return !v
}

func when(v bool) bool {
    return v
}
```

I'd figure the compiler optimizes these away and they would effectively just help with readability.  I'm unsure about the latter `when()` though, but just trying to think ways to make reviewing easier :slightly_smiling_face:

cmd/cluster/scale/defaulting/scaling.go
```diff
@@ -0,0 +1,131 @@
...
+				func() bool { return !s.desiredScalingMaxChanged },
+				func() bool { return s.desiredNumWorkersChanged },
+			},
+			Defaulting: func() {
+				scaling.Min = s.desiredNumWorkers
+				scaling.Max = s.desiredNumWorkers
```
tuommaki: Just so that I understand: When this defaulting is in `gsctl`, it should be always intentional by user to pin number of workers to static number even if there was auto-scaling range present before this?

cmd/cluster/scale/defaulting/scaling.go
```diff
@@ -0,0 +1,131 @@
...
+		{
+			Conditions: []func() bool{
+				func() bool { return !s.autoScalingEnabled },
+				func() bool { return s.desiredScalingMaxChanged },
+			},
+			Defaulting: func() {
+				scaling.Min = s.desiredScalingMax
+				scaling.Max = s.desiredScalingMax
+			},
+		},
```
tuommaki: Does validation catch situation where both scaling-min or scaling-max and number of workers is given? Otherwise if I understand this correctly, this defaulting makes the cluster then pinned at number of workers and that is not always what the user intended. The user could e.g. misunderstand how this works and try to set scaling limits and then set present "desired number of workers" to some specific values - which is impossible.

cmd/cluster/scale/defaulting/scaling_test.go
```diff
@@ -0,0 +1,140 @@
...
+			CurrentScalingMin:        3,
+			DesiredNumWorkers:        0,
+			DesiredNumWorkersChanged: false,
+			DesiredScalingMax:        0,
+			DesiredScalingMaxChanged: false,
+			DesiredScalingMin:        5,
+			DesiredScalingMinChanged: true,
+			ExpectedScaling: request.Scaling{
+				Min: 5,
+				Max: 3,
```
tuommaki: I assume this (min > max) is only to test that the defaulting logic touches only relevant field?

cmd/cluster/scale/defaulting/scaling.go
xh3b4sd: We do not have optimization issues, especially not in the command line clients. Not on this level. I like the idea with `not`. This is a good idea. Not sure about `when`. I figured it is pretty hard to read with longer variable names in camel case. In the future I would like to have more structured structures so to speak. `desiredScalingMinChanged` would then be `desired.Scaling.Min.Changed`.

cmd/cluster/scale/defaulting/scaling.go
xh3b4sd: Maybe? Not sure I can answer that. We have quite a complexity explosion with the whole flag mess. We have to get better with this in the future.

cmd/cluster/scale/defaulting/scaling.go
xh3b4sd: Same as above, I do not know. When there is a case you have in mind let's test and verify. There are a couple of confirmations so I am not sure how misleading all of this can be at the end.

cmd/cluster/scale/defaulting/scaling_test.go
xh3b4sd: Right, defaulting works. Validation is a different thing, which does not apply here.

cmd/cluster/scale/defaulting/scaling.go
xh3b4sd: I added `not`.

PR #344, opened 2019-02-07
Make 'delete cluster' accept cluster ID as positional argument

marians:
Fixes https://github.com/giantswarm/gsctl/issues/94 for UX alignment reasons

The old form is still accepted.


xh3b4sd (approved):
I fell for this several times. Cool thing.

pipo02mix (approved):
like it!

marians:
I was really getting annoyed, too, during latests tests. It took only a few more words from @MarcelMue to finally trigger the change.

PR #345, opened 2019-02-11
Use current workers count for confirmation instead of desired

MarcelMue:
Use actual worker count for confirmation instead of desired count.

### DIFF
--- a/commands/scale_cluster.go
+++ b/commands/scale_cluster.go
@@ -89,15 +89,15 @@ func init() {
 }
 
 // confirmScaleCluster asks the user for confirmation for scaling actions.
-func confirmScaleCluster(args scaleClusterArguments, maxBefore int64, minBefore int64, desiredCapacity int64) error {
-	if desiredCapacity > args.workersMax && args.workersMax == args.workersMin {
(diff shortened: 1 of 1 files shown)


commands/scale_cluster.go
```diff
@@ -316,7 +330,7 @@ func scaleClusterRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
 
 	// Ask for confirmation for the scaling action.
 	if !cmdForce {
-		err = confirmScaleCluster(args, currentScalingMax, currentScalingMin, desiredCapacity)
+		err = confirmScaleCluster(args, currentScalingMax, currentScalingMin, statusWorkers)
```
xh3b4sd: I think the wording with current and desired is a good thing. Also having the same semantics when providing and argument and using it within a function. Now `statusWorkers` is given and in the function it is called `currentWorkers`. That is all pretty confusing and does not make it better.

xh3b4sd:
This case is where I expect the confirmation to appear, but it does not. 

```
$ gsctl scale cluster 7obj3 --workers-min 2
```

This one is good. 

```
$ gsctl scale cluster 7obj3 --workers-max 2
Number of worker nodes invalid
Node count flag --workers-min must not be higher than --workers-max.
```

This one is good. 

```
$ gsctl scale cluster 7obj3 --workers-min 2 --workers-max 2
The cluster currently has 3 worker nodes running.
Do you want to pin the number of worker nodes to 2? [y/n]:
```

/cc @marians

PR #346, opened 2019-02-11
Fix delete cluster flag change to positional argument

marians:
- Handle 404 case
- Show the cluster ID in success message when using legacy `-c`/`--cluster` flag
- Actually deprecate the flag


pipo02mix (approved):
LGTM

PR #347, opened 2019-02-14
Dependency update

marians:
This activates pruning for non Go files and updates all dependencies.


marians:
For dependency updating I would normally not request for reviews. Just want to have confirmation that the pruning is fine.

xh3b4sd:
Should be fine. What are your concerns?

marians:
Nothing specific. Thanks!

PR #351, opened 2019-03-13
Using KubeConfig library 

tomahawk28:
- Using https://github.com/giantswarm/kubeconfig/ when creating kubeconfig files. 
- Nothing changes in UI/UX


commands/create_kubeconfig.go
```diff
@@ -452,43 +406,37 @@ func createKubeconfig(args createKubeconfigArguments) (createKubeconfigResult, e
...
+			if err != nil {
+				return result, microerror.Mask(err)
+			}
 
-		// create YAML
-		yamlBytes, err := yaml.Marshal(&kubeconfig)
-		if err != nil {
-			return result, microerror.Mask(err)
+			kcc := kubeconfig.Config{
+				Logger: microloggertest.New(),
```
tomahawk28: Using a test structure, since we do not use it inside the library.

commands/create_kubeconfig.go
```diff
@@ -452,43 +406,37 @@ func createKubeconfig(args createKubeconfigArguments) (createKubeconfigResult, e
...
+				K8sClient: fake.NewSimpleClientset(),
+			}
+
+			kc, err := kubeconfig.New(kcc)
+			if err != nil {
+				return result, microerror.Mask(err)
+			}
+			yamlBytes, err = kc.NewKubeConfigForRESTConfig(nil, restConfig, fmt.Sprintf("giantswarm-%s", args.clusterID), "")
```
tomahawk28: context should nil? or background?

commands/create_kubeconfig.go
```diff
@@ -11,9 +10,12 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/giantswarm/gsclientgen/models"
+	"github.com/giantswarm/kubeconfig"
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"github.com/giantswarm/operatorkit/client/k8srestconfig"
 	"github.com/spf13/cobra"
-	yaml "gopkg.in/yaml.v2"
+	"k8s.io/client-go/kubernetes/fake"
```
rossf7: Hmm this is a bit tricky. We should avoid using the test libraries in gsctl isn't nice.

I think in the kubeconfig library we should relax the validation so the logger and k8s client are not required. As they are not used for `NewKubeConfigForRESTConfig`.

commands/create_kubeconfig.go
rossf7: We should provide a background context. Ideally this should be at the entrypoint to the command. Later we might need to use the context for calling other libraries.

rossf7 (commented):
Its nice we use the kubeconfig library here but some changes are needed.

@marians Is there something Ji needs to do about the CI error due to the test coverage? I have no clue about that.

commands/create_kubeconfig.go
tomahawk28: Or what do you think that we could create another struct in `spec.go` and implement two methods as `NewKubeConfigForRestConfig`? or we could implement it as class method (don't know equivalent word for golang :( )

rossf7: Sorry about the delayed reply.

 I think we should simplify the interface. We only need the context and logger for the `ForApp` method because it fetches the secret and we should have logging there.


So it would look like this.

```
type Interface interface {
	// NewRESTConfigForApp returns a Kubernetes REST Config for the cluster configured
	// in the kubeconfig section of the app CR.
	NewRESTConfigForApp(ctx context.Context, app v1alpha1.App) (*rest.Config, error)
	// NewKubeConfigForRESTConfig returns a kubeConfig bytes for the given REST Config.
	NewKubeConfigForRESTConfig(config *rest.Config, clusterName, namespace string) ([]byte, error)
	// NewRESTConfigForKubeConfig returns a REST Config for the given KubeConfigValue.
	NewRESTConfigForKubeConfig(kubeConfig []byte) (*rest.Config, error)
}
```

I think that's simpler and much easier to use in gsctl.

For changing the interface in golang AFAIK you can't have 2 methods with the same name and signature. Its more restrictive than other languages.

WDYT?

tomahawk28: That looks good, but I'm not sure we can set the loose restriction on `k8sclient` and `logger` when the same `kubeconfig` instance actually could use other `ForApp` methods even though it shouldn't

rossf7: Best option I can see for now is doing the config checks in the `ForApp` method. Maybe with a `validate` function as more methods might need it in future.

marians:
Coveralls was meant to be informational only. Either add a unit test to increase the coverage (not serious) or ignore it.

tomahawk28:
Address reviews, @rossf7. @marians PTAL :)

commands/create_kubeconfig.go
```diff
@@ -452,43 +407,28 @@ func createKubeconfig(args createKubeconfigArguments) (createKubeconfigResult, e
...
-						ClientKeyData:         base64.StdEncoding.EncodeToString([]byte(response.Payload.ClientKeyData)),
-					},
+		var yamlBytes []byte
+		{
+			c := k8srestconfig.Config{
+				Logger: microloggertest.New(),
+
+				Address:   result.apiEndpoint,
+				InCluster: false,
+				TLS: k8srestconfig.TLSClientConfig{
```
rossf7:
```suggestion
				TLS: k8srestconfig.ConfigTLS{
```

This got renamed. You'll need to do a vendor update for operatorkit.

commands/create_kubeconfig.go
```diff
@@ -452,43 +407,28 @@ func createKubeconfig(args createKubeconfigArguments) (createKubeconfigResult, e
...
-				{
-					Name: "giantswarm-" + args.clusterID + "-user",
-					User: KubeconfigUserKeyPair{
-						ClientCertificateData: base64.StdEncoding.EncodeToString([]byte(response.Payload.ClientCertificateData)),
-						ClientKeyData:         base64.StdEncoding.EncodeToString([]byte(response.Payload.ClientKeyData)),
-					},
+		var yamlBytes []byte
+		{
+			c := k8srestconfig.Config{
+				Logger: microloggertest.New(),
```
rossf7: We should only use the `microloggertest` package in unit tests. Lets just add a regular micrologger.

commands/create_kubeconfig.go
```diff
@@ -452,43 +407,28 @@ func createKubeconfig(args createKubeconfigArguments) (createKubeconfigResult, e
...
+			restConfig, err := k8srestconfig.New(c)
+			if err != nil {
+				return result, microerror.Mask(err)
+			}
 
-		// create YAML
-		yamlBytes, err := yaml.Marshal(&kubeconfig)
-		if err != nil {
-			return result, microerror.Mask(err)
+			yamlBytes, err = kubeconfig.NewKubeConfigForRESTConfig(context.Background(), restConfig, fmt.Sprintf("giantswarm-%s", args.clusterID), "")
```
rossf7: I don't like defining the context here. We should create it at the entrypoint to the command.

I think it should be here. https://github.com/giantswarm/gsctl/blob/786cb6418360a4d52dbb8ff57e18559a29795a0c/commands/create_kubeconfig.go#L163

rossf7 (commented):
Good to remove all the structs but a few changes needed.

commands/create_kubeconfig.go
tomahawk28: Thanks, didn't know that

rossf7: I posted it in #news-dev because it was a breaking change albeit a small one. 

https://gigantic.slack.com/archives/C04TGHDEF/p1552555043000200

It's good to keep up to date with the news channels for stuff like this.

commands/create_kubeconfig.go
tomahawk28: That's in `createKubeconfigPreRunOutput`, I don't think it's nice passing around the context which would use just only one time.

rossf7: The context should be created at the entry point and flow through all methods. Like here in `opsctl`. If each function call has its own context package it doesn't help.

https://github.com/giantswarm/opsctl/blob/a948553d66cd789673d6671d2e9635e6f20c7ea8/command/update/installation/command.go#L111

tomahawk28: @rossf7 I can create `context` in `createKubeconfigRunOutput` instead, WDYT?

rossf7: Yes creating it there is good. Thx!

tomahawk28:
@giantswarm/sig-ux @giantswarm/team-batman PTAL :)

rossf7 (approved):
LGTM when tested.

tomahawk28:
Tested ```./gsctl create kubeconfig -c w7utg --self-contained ./self/k --certificate-organizations=system:masters```
Works with kubectl.

PR #356, opened 2019-04-26
Capabilities package to simplify decision making around availability of features, persisting of provider info in config

marians:
To be merged into https://github.com/giantswarm/gsctl/pull/355

This PR

- Adds the `provider` key to the config to persist the provider together with an endpoint, and also adds an `EnsureProvider` function to be called in every command that needs this info.
- introduces the `capabilities` package which makes it relatively easy to encode and query the capabilities based on provider and tenant cluster release version.
- Applies both to the `scale cluster` command.

### DIFF
--- a/capabilities/capabilities.go
+++ b/capabilities/capabilities.go
@@ -17,7 +17,7 @@ var (
 		},
 	}
 
-	// AvailabilityZones si the capability to spread the worker nodes of a tenant
+	// AvailabilityZones is the capability to spread the worker nodes of a tenant
 	// cluster over multiple availability zones.
 	AvailabilityZones = CapabilityDefinition{
 		Name: "AvailabilityZones",
--- a/capabilities/capabilities.go
+++ b/capabilities/capabilities.go
@@ -6,13 +6,25 @@ import (
 )
 
 var (
-	// Autoscaling si the capability to scale tenant clusters automatically.
+	// Autoscaling is the capability to scale tenant clusters automatically.
 	Autoscaling = CapabilityDefinition{
 		Name: "Autoscaling",
 		RequiredReleasePerProvider: []ReleaseProviderPair{
 			ReleaseProviderPair{
 				Provider:       "aws",
 				ReleaseVersion: semver.MustParse("6.3"),
+			},
+		},
+	}
+
+	// AvailabilityZones si the capability to spread the worker nodes of a tenant
+	// cluster over multiple availability zones.
+	AvailabilityZones = CapabilityDefinition{
+		Name: "AvailabilityZones",
+		RequiredReleasePerProvider: []ReleaseProviderPair{
+			ReleaseProviderPair{
+				Provider:       "aws",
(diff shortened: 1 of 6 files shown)


marians:
FYI: I am thinking about cherry-picking this into master, because it does not actually require V5 API calls yet, so it's not depending on node pools.

capabilities/capabilities.go
```diff
@@ -0,0 +1,95 @@
...
+		Name: "Autoscaling",
+		RequiredReleasePerProvider: []ReleaseProviderPair{
+			ReleaseProviderPair{
+				Provider:       "aws",
+				ReleaseVersion: semver.MustParse("6.3"),
+			},
+		},
+	}
+
+	// AvailabilityZones si the capability to spread the worker nodes of a tenant
```
pipo02mix:
```suggestion
	// AvailabilityZones is the capability to spread the worker nodes of a tenant
```

pipo02mix (approved):
LGTM but I would wait for someone validate fmt

capabilities/capabilities.go
```diff
@@ -0,0 +1,95 @@
...
+			cap = append(cap, capability)
+		}
+	}
+
+	return cap, nil
+}
+
+// HasCapability returns true if the current context (provider, release) provides
+// the given capabililty.
+func HasCapability(provider string, releaseVersion *semver.Version, capability CapabilityDefinition) bool {
```
tuommaki:
```suggestion
func HasCapability(provider, releaseVersion string, capability CapabilityDefinition) bool {
```

I'd maybe accept the release version here as a string and perform the parsing inside. Two reasons:
* It doesn't tie this API to specific library (`semver`).
* By looking at code below it seems a bit ugly to do parsing before call to this.

I'm not fully sure which approach is better: Perform error logging & `os.Exit(1)` in here or return the error, but when release version is something that's quite deeply built into our product, I'd assume it's always an unrecoverable error when it cannot be parsed and therefore just terminating here would probably be correct.

marians: Thanks for the suggestion. I'll do the semver parsing inside of capabilities.

I wouldn't panic or exit in the `capabilities` package, but return the error instead. Treating that should be the responsibility of the command packages.

marians: Done

capabilities/capabilities_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+type testInput struct {
+	Provider       string
+	ReleaseVersion string
+}
+
+var capabilityTests = []struct {
+	in  testInput
+	out []CapabilityDefinition
+}{
```
tuommaki: Maybe add one or two tests with `ReleaseVersion` that doesn't parse (to verify that error is returned).

tuommaki (approved):
LGTM; minor note about possible addition of one or two tests.

capabilities/capabilities_test.go
marians: Done

PR #357, opened 2019-05-03
Remove one of our semver dependencies

marians:
We had two overlapping dependencies in use:

- `github.com/coreos/go-semver`
- `github.com/Masterminds/semver`

This PR removes the first.


pipo02mix (approved):

oponder (approved):
LGTM, though a util package is not go idiomatic: https://blog.golang.org/package-names

marians:
@oponder Correct, but as I introduced that mistake long before this PR, I'll ignore that now.

PR #360, opened 2019-05-13
Refactoring command tree

marians:
The `commands` package has gotten slightly out of hand.

This PR aims for

- a tree structure in the `commands` package, resembling at least the first two command levels
- better separation of concerns

The intended tree structure so far is only applied to the `show` command, to avoid making the PR too big.

Among the changes are

- new package `flags` for command flag variables that are used in multiple commands
- new package `commands/errors` for errors that occur in command validation or execution (including user input)
- handling of common errors split into `commands/errors` and `client` to get rid of cyclic imports
- Now uses Golang 1.12 in CI and in the Makefile
- Removes Codecov integration, as coveralls turns out to work at least as good and is less chatty in PRs.

In addition, CI was failing to run tests in many cases because of what seems to be memory problems. Upgraded to a higher resource class.

Imports structure:

image


client/client.go
```diff
@@ -389,6 +403,27 @@ func (w *WrapperV2) GetCluster(clusterID string, p *AuxiliaryParams) (*clusters.
...
+		return "", clienterror.New(err)
+	}
+
+	if len(response.Payload) == 1 {
+		return response.Payload[0].ID, nil
+	}
+
+	return "", nil
+}
+
```
marians: Moved from the `config` package to `client`

client/error.go
```diff
@@ -32,12 +40,45 @@ func IsEndpointNotSpecifiedError(err error) bool {
...
+
+	if headline == "" {
+		return
+	}
+
+	fmt.Println(color.RedString(headline))
+	if subtext != "" {
+		fmt.Println(subtext)
+	}
+	os.Exit(1)
```
marians: I had to split the big `errors.HandleCommonErrors` due to cyclic imports. Now there is new function for printing common errors occurring in the `client` package.

tobbbles: Seems like a sensible change - keeps everything nice and localised to the package

commands/create_cluster.go
```diff
@@ -237,36 +240,37 @@ func createClusterExecutionOutput(cmd *cobra.Command, args []string) {
 
 	result, err := addCluster(aca)
 	if err != nil {
-		handleCommonErrors(err)
+		errors.HandleCommonErrors(err)
+		client.HandleErrors(err)
```
tobbbles: Could these be moved inside the switch statement?

commands/create_cluster_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/giantswarm/gsctl/commands/errors"
+	"github.com/giantswarm/gsctl/flags"
```
tobbbles: Internal import to be ordered below 3rd party

commands/errors/handling.go
```diff
@@ -0,0 +1,89 @@
...
+	} else if convertedErr, ok := err.(*clienterror.APIError); ok {
+		headline = convertedErr.ErrorMessage
+		subtext = convertedErr.ErrorDetails
+	} else {
+		// legacy client error handling
+		switch {
+		case oidc.IsAuthorizationError(err):
+			headline = "Unauthorized"
+			subtext = "Something went wrong during a OIDC operation: " + err.Error() + "\n"
+			subtext += "Please try logging in again."
```
tobbbles: I'd like to see use of `fmt.Sprintf` here to keep this tidy in one line. Optional change as it's just cleaning up legacy bits

marians: Done

commands/create_cluster_test.go
marians: Changed

commands/create_cluster.go
marians: I wouldn't know how. They are meant to print a message in case a known error happened and exit.

marians:
@tobbbles I consider this done now. If you could glance over the added changes, I'd be more than happy.

commands/root.go
```diff
@@ -6,8 +6,19 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/giantswarm/gsctl/client"
+	"github.com/giantswarm/gsctl/commands/create"
+	deletecmd "github.com/giantswarm/gsctl/commands/delete"
```
tobbbles: Is this import rename required?

tobbbles: Yes it is, `delete` is a reserved key word :man_facepalming:

confirm/confirm.go
```diff
@@ -0,0 +1,36 @@
...
+	"strings"
+
+	"github.com/fatih/color"
+)
+
+// AskForConfirmation asks the user for confirmation. A user must type in "yes" or "no" and
+// then press enter. It has fuzzy matching, so "y", "Y", "yes", "YES", and "Yes" all count as
+// confirmations. If the input is not recognized, it will ask again. The function does not return
+// until it gets a valid response from the user.
+func AskForConfirmation(s string) bool {
```
tobbbles: This is a small nit-pick, but when reading confim.AskForConfirmation it stutters a little, I'd personally like to see confirm.Ask here, but it's down to you!

```suggestion
func Ask(s string) bool {
```

confirm/confirm.go
```diff
@@ -0,0 +1,36 @@
...
+		response, err := reader.ReadString('\n')
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		response = strings.ToLower(strings.TrimSpace(response))
+
+		if response == "y" || response == "yes" {
+			return true
+		} else if response == "n" || response == "no" {
```
tobbbles: This could be cleaned up in a case, see below


```go
switch strings.ToLower(strings.TrimSpace(response)) {
	case "y", "yes":
		return true
	case "n", "no":
		return false
	}
}
```

confirm/confirm.go
marians: Done

confirm/confirm.go
marians: Done

PR #363, opened 2019-05-14
Command refactoring: datele, info, login, logout, scale, upgrade, update, version

marians:


commands/info/command.go
```diff
@@ -64,11 +67,23 @@ type infoResult struct {
...
+// printValidation prints if there is anything missing from user input or config.
+func printValidation(cmd *cobra.Command, extraArgs []string) {
+	args := defaultInfoArguments()
+	err := validatePreconditions(args)
+
+	if err != nil {
+		errors.HandleCommonErrors(err)
+	}
+}
+
```
marians: Adding `validatePreconditions` and `printValidation` as they are in place normally, to explicitly remind us here that this command in fact does not have any preconditions to be met.

PR #367, opened 2019-05-28
Use spf13/afero for file system access

marians:
This PR introduces https://github.com/spf13/afero for filesystem access.

In tests, this allows to use a memory mapping file system in most cases, which allows that tests can run in isolation from each other and be faster.


tobbbles (approved):
Sensible looking change, nothing appears to be an issues 🉑

tuommaki (approved):
Glanced over. Seems fine when it works for you :+1:

marians:
Thanks, gentlemen!

PR #371, opened 2019-07-10
Use shared gscliauth package

schjan:
What changed on gsctl side?
- Code was moved to giantswarm/gscliauth repository. (Because of that coverage decreased)
- Gopkg.toml was changed accordingly
- spf13/afero was set to a fixed version so that it is compatible to giantswarm/gscliauth and giantswarm/opsctl

_Part of giantswarm/giantswarm#6145 story_


MarcelMue (approved):
LGTM - Please add sig-ux as reviewers as they currently have a big stake in gsctl.

marians:
I'll give this a test today.

So the `spf13/afero` version pinning is forced by opsctl? What's the reason that opsctl needs a specific version?

schjan:
Pinning to `master` in one project and to a version tag in a dependency is not working with `dep`. So we can either go `master` or fixed version.
I decided for fixed version (everything < 2) so we don't have weird dependency issues on a major version change.

marians (approved):
Works for me

PR #372, opened 2019-07-11
Add v5 support, node pools

marians:

## Changes

- Let the `gsctl show cluster` command also support v5 clusters which contain node pools.
- Add the new hidden command `gsctl list nodepools`.
- Add the new hidden command `gsctl show nodepool`.

This branch also contains support for embedded AWS instance type information. This may be removed again later once the API's info endpoint delivers instance type details.


client/clienterror/clienterror.go
```diff
@@ -222,6 +223,65 @@ func New(err error) *APIError {
 		}
 	}
 
+	// get v4 cluster status
+	if getClusterStatusDefaultErr, ok := err.(*clusters.GetClusterStatusDefault); ok {
```
MarcelMue: We talked about refactoring this before IIRC - is there a better way to do this?

client/client.go
```diff
@@ -403,6 +404,38 @@ func (w *WrapperV2) GetCluster(clusterID string, p *AuxiliaryParams) (*clusters.
 	return response, nil
 }
 
+// GetClusterV5 fetches details on a V5 cluster.
+func (w *WrapperV2) GetClusterV5(clusterID string, p *AuxiliaryParams) (*clusters.GetClusterV5OK, error) {
```
MarcelMue: Would it make more sense to have a package separation here like we do in operators instead of renaming our functions?

commands/show/cluster/command_test.go
```diff
@@ -49,22 +61,146 @@ func TestShowAWSCluster(t *testing.T) {
 		clusterID:   "cluster-id",
 		scheme:      "giantswarm",
 		authToken:   "my-token",
+		verbose:     true,
```
MarcelMue: Why verbose?

nodespec/aws.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/giantswarm/microerror"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	// specYAML is the raw data on all necessary AWS instance types taken from
+	// https://github.com/giantswarm/installations/blob/master/default-draughtsman-configmap-values.yaml
+	// Warning: YAML in Golang is super fragile. There must not be any tabs in this string, otherwise
+	// the marshalling will fail. However we will likely detect this in CI when running tests.
+	awsInstanceTypesYAML = `
```
MarcelMue: So here is my confusion - in the test you wrote above there is memory and storage provided. This should still work for all providers because it comes from `cluster-service` like before. What am I missing?

Gopkg.toml
```diff
@@ -39,7 +39,7 @@
   version = "=2.0.2"
 
 [[constraint]]
-  branch = "master"
+  branch = "node-pools"
```
marians: This will be reverted in the end

client/client.go
marians: Not sure, as I don't develop on operators

client/clienterror/clienterror.go
marians: I don't know if there is.

commands/show/cluster/command_test.go
marians: I enabled it because I wanted to actually see what's happening in the test. This way the test actually covers more code lines.

PR #373, opened 2019-07-12
Add 'list nodepools' command

marians:
Towards https://github.com/giantswarm/giantswarm/issues/6447, to be merged into https://github.com/giantswarm/gsctl/pull/372

This PR adds the new `gsctl list nodepools <cluster-id>` command to list node pools of a cluster.

## Preview

```nohighlight
ID     NAME                    AZ     INSTANCE TYPE  NODES MIN/MAX  NODES DESIRED  NODES READY  CPUS  RAM (GB)
6feel  Application servers     A,B,C  p3.2xlarge     3/15           10             9            72    549.0
a7rc4  Batch number crunching  D      p3.8xlarge     2/5            4              4            128   976.0
```

Screenshot

image

### DIFF
--- a/commands/list/nodepools/command.go
+++ b/commands/list/nodepools/command.go
@@ -1,0 +1,119 @@
+// Package nodepools implements the 'list organizations' sub-command.
+package nodepools
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/fatih/color"
+	"github.com/giantswarm/gsclientgen/client/nodepools"
+	"github.com/giantswarm/microerror"
+	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/gsctl/client"
+	"github.com/giantswarm/gsctl/client/clienterror"
+	"github.com/giantswarm/gsctl/commands/errors"
+	"github.com/giantswarm/gsctl/config"
+	"github.com/giantswarm/gsctl/flags"
+)
+
+var (
+	// Command performs the "list organizations" function
+	Command = &cobra.Command{
+		Use:     "nodepools <cluster-id>",
+		Aliases: []string{"nps", "np"},
+
+		// Args: cobra.ExactArgs(1) guarantees that cobra will fail if no positional argument is given.
+		Args:  cobra.ExactArgs(1),
+		Short: "List node pools",
+		Long: `Prints a list of the node pools of a cluster.
+
+The result will be a table of all node pools of a specific cluster with some details.
+
+To see all available details for a cluster, use 'gsctl show nodepool <cluster-id>/<nodepool-id>'.
+
+To list all clusters you have access to, use 'gsctl list clusters'.`,
+		PreRun: printValidation,
+		Run:    printResult,
+	}
+)
+
+const activityName = "list-nodepools"
+
+type arguments struct {
+	apiEndpoint string
+	authToken   string
+	scheme      string
+	clusterID   string
+}
+
+// defaultArgs creates arguments based on command line flags and config.
+func defaultArgs(cmdLineArgs []string) arguments {
+	endpoint := config.Config.ChooseEndpoint(flags.CmdAPIEndpoint)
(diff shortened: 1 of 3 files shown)


commands/list/nodepools/command.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+		} else {
+			fmt.Println(color.RedString("Error: %s", err.Error()))
+		}
+		os.Exit(1)
+	}
+
+	awsInfo, err := nodespec.NewAWS()
+	if err != nil {
+		fmt.Println(color.RedString("Error: Cannot provide info on AWS instance types. Details: %s", err))
```
xh3b4sd: As a general rule of thumb, errors should always be propagated back to the caller, so they can be handled there if desired. A program flow should be simple. A program starts, a program ends. A program should end where it started. This here now prints something and still execution will proceed. From here more will go wrong probably.

commands/list/nodepools/command.go
```diff
@@ -0,0 +1,206 @@
...
+
+		if clientErr, ok := err.(*clienterror.APIError); ok {
+			fmt.Println(color.RedString(clientErr.ErrorMessage))
+			if clientErr.ErrorDetails != "" {
+				fmt.Println(clientErr.ErrorDetails)
+			}
+		} else {
+			fmt.Println(color.RedString("Error: %s", err.Error()))
+		}
+		os.Exit(1)
```
xh3b4sd: Same as for the error handling. Best practise is to not stop programs somewhere. There is at best one place where the program starts and terminates. That is highly beneficial to code maintenance where different people work on the same project.

commands/list/nodepools/command.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/giantswarm/gsctl/commands/errors"
+	"github.com/giantswarm/gsctl/config"
+	"github.com/giantswarm/gsctl/flags"
+	"github.com/giantswarm/gsctl/nodespec"
+)
+
+var (
+	// Command performs the "list organizations" function
+	Command = &cobra.Command{
+		Use:     "nodepools <cluster-id>",
```
tobbbles: When listing keypairs, we pass the cluster ID in as a flag (https://github.com/giantswarm/gsctl/blob/master/commands/list/keypairs/command.go#L61), I personally feel that's cleaner and we should stay consistent here

marians: We are actually gradually moving towards positional arguments everywhere in gsctl when an option is mandatory.

commands/list/nodepools/command.go
marians: That's something we should then discuss outside of this PR, as it is done this way in every single command.

commands/list/nodepools/command.go
marians: The intend here was to report this error, but treat it as non-breaking. Anyway, I think I will change that to avoid this special treatment here.

commands/list/nodepools/command.go
xh3b4sd: I brought it up every once in a while but nothing happened really. Somebody needs to feel responsible and drive it and I cannot really do more than bring up the topic again.

commands/list/nodepools/command.go
tobbbles: OK, then that makes sense

commands/list/nodepools/command.go
marians: This is changed now. I gathered all data collection in a common function, moving this part out of the `printResult` function.

marians:
@tobbbles @xh3b4sd Can you re-check please?

commands/list/nodepools/command.go
xh3b4sd: For the record. We talked about this and want to try to drive certain efforts in the scope of special working hours (SWH). I think this is a pretty good idea to realize more software design related standards.

PR #374, opened 2019-07-16
Add 'show nodepool' command

marians:
To be merged into https://github.com/giantswarm/gsctl/pull/372

Part of https://github.com/giantswarm/giantswarm/issues/6447

This PR adds the `gsctl show nodepool <cluster-id>/<nodepool-id>` command.

### Preview

```nohighlight
$ gsctl show nodepool m0ckd/a7rc4
ID:                    a7rc4
Name:                  Batch number crunching
Node instance type:    p3.8xlarge - 244 GB RAM, 32 CPUs each
Availability zones:    D
Node scaling:          Autoscaling between 2 and 5
Nodes desired:         4
Nodes in state Ready:  4
CPUs:                  128
RAM:                   976 GB
```

### DIFF
--- a/commands/create/keypair/command.go
+++ b/commands/create/keypair/command.go
@@ -18,6 +18,7 @@ import (
 	"github.com/giantswarm/gsctl/commands/errors"
 	"github.com/giantswarm/gsctl/config"
 	"github.com/giantswarm/gsctl/flags"
+	"github.com/giantswarm/gsctl/formatting"
 	"github.com/giantswarm/gsctl/util"
 )
 
@@ -206,7 +207,7 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
 
 	// Success output
 	msg := fmt.Sprintf("New key pair created with ID %s and expiry of %v",
-		util.Truncate(util.CleanKeypairID(result.id), 10, true),
+		util.Truncate(formatting.CleanKeypairID(result.id), 10, true),
 		util.DurationPhrase(int(result.ttlHours)))
 	fmt.Println(color.GreenString(msg))
 
--- a/commands/show/nodepool/command.go
+++ b/commands/show/nodepool/command.go
@@ -22,9 +22,10 @@ import (
 var (
 	// ShowNodepoolCommand is the cobra command for 'gsctl show nodepool'
 	ShowNodepoolCommand = &cobra.Command{
-		Hidden:  true,
-		Use:     "nodepool <cluster-id>/<nodepool-id>",
-		Aliases: []string{"np"},
+		DisableFlagsInUseLine: true,
+		Hidden:                true,
+		Use:                   "nodepool <cluster-id>/<nodepool-id>",
+		Aliases:               []string{"np"},
 		// Args: cobra.ExactArgs(1) guarantees that cobra will fail if no positional argument is given.
 		Args:  cobra.ExactArgs(1),
 		Short: "Show node pool details",
--- a/commands/show/nodepool/command.go
+++ b/commands/show/nodepool/command.go
@@ -1,0 +1,176 @@
+// Package nodepool implements the 'show nodepool' command.
+package nodepool
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/giantswarm/columnize"
+	"github.com/giantswarm/gsclientgen/models"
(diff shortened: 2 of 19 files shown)


formatting/key_pair_id_test.go
```diff
@@ -0,0 +1,21 @@
+package formatting
+
+import "testing"
+
+type KeypairIDTest struct {
+	input  string
+	output string
+}
+
+var keypairIDTests = []KeypairIDTest{
```
xh3b4sd: We have https://github.com/giantswarm/fmt/blob/master/go/unit_tests.md. You have no naming issues when you for instance put the structure inline.

commands/show/nodepool/command.go
```diff
@@ -0,0 +1,178 @@
...
+
+var (
+	// ShowNodepoolCommand is the cobra command for 'gsctl show nodepool'
+	ShowNodepoolCommand = &cobra.Command{
+		DisableFlagsInUseLine: true,
+		Hidden:                true,
+		Use:                   "nodepool <cluster-id>/<nodepool-id>",
+		Aliases:               []string{"np"},
+		// Args: cobra.ExactArgs(1) guarantees that cobra will fail if no positional argument is given.
+		Args:  cobra.ExactArgs(1),
```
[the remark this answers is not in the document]
marians: Good enough, yes.

```nohighlight
Error: accepts 1 arg(s), received 0
Usage:
  gsctl show nodepool <cluster-id>/<nodepool-id>

...
```

commands/create/keypair/command.go
```diff
@@ -206,7 +207,7 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
 
 	// Success output
 	msg := fmt.Sprintf("New key pair created with ID %s and expiry of %v",
-		util.Truncate(util.CleanKeypairID(result.id), 10, true),
+		util.Truncate(formatting.CleanKeypairID(result.id), 10, true),
```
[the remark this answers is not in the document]
marians: Yet another case of "let me fix this while I'm here". :/

formatting/availability_zones.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"sort"
+	"strings"
+)
+
+// AvailabilityZonesList returns the list of availability zones
+// as one string consisting of uppercase letters only, separated by comma.
+// Example: "A,B,C".
+func AvailabilityZonesList(az []string) string {
```
tobbbles: Do we forsee ever having an `az` list of, for example, `eu-west1a, eu-west1b, eu-west2a`? And if so how do we expect to handle this?

formatting/availability_zones_test.go
```diff
@@ -0,0 +1,20 @@
+package formatting
+
+import "testing"
+
+var flagtests = []struct {
```
tobbbles: Would be much nicer to have the test tables in the standard way of unit tests defined [here](https://github.com/giantswarm/fmt/blob/master/go/unit_tests.md) 👍

formatting/availability_zones.go
marians: Having both `eu-west1` and `eu-west2` in the same list is no considered currently, as these would be different regions, which is not supported by our product.

formatting/availability_zones_test.go
marians: Done

formatting/key_pair_id_test.go
marians: Done

commands/create/keypair/command.go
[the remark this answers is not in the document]
xh3b4sd: That is ok. Now we maybe have an idea of a trigger thought to separate changes better in the future. You can work that out. 👍

formatting/key_pair_id_test.go
```diff
@@ -0,0 +1,26 @@
+package formatting
+
+import (
+	"strconv"
+	"testing"
+)
+
+type KeypairIDTest struct {
```
xh3b4sd: Not sure if the structure is necessary. In the other test you inlined the struct. Consistency and `fmt` alignment would be appreciated.

marians: Removed

marians:
Thanks for the comments. One by one I figure out all the details covered by fmt :)

PR #377, opened 2019-07-19
Add delete app command.

oponder:
Adds the `gsctl delete app <appname> -c <clusterid>` command.

Towards: https://github.com/giantswarm/giantswarm/issues/6518


commands/delete/app/command_test.go
```diff
@@ -0,0 +1,111 @@
...
+		w.WriteHeader(http.StatusAccepted)
+		w.Write([]byte(`{"code": "RESOURCE_DELETED", "message": ""}`))
+	}))
+	defer mockServer.Close()
+
+	testutils.CaptureOutput(func() {
+		Command.SetArgs([]string{"--force", "--endpoint", mockServer.URL})
+		Command.Execute()
+	})
+}
```
marians: Can we add some notice here and also in commands/delete/cluster.TestCommandExecution to remind ourselves that these will print an error because the global flag `--endpoint` is not known? We should fix this at some point.

marians (approved):
LGTM, nothing blocking.

PR #378, opened 2019-07-19
Add reusable EnsureProviderInConfig function in client

marians:
This adds a `EnsureProviderInConfig` function to the client package, as discussed in a hangout, to have it re-usable form other commands.

### DIFF
--- a/commands/scale/cluster/command.go
+++ b/commands/scale/cluster/command.go
@@ -173,12 +173,6 @@ func getClusterStatus(clusterID, activityName string) (*client.ClusterStatus, er
 	auxParams := clientV2.DefaultAuxiliaryParams()
 	auxParams.ActivityName = activityName
 
-	// Make sure we have provider info in the current endpoint
-	err = clientV2.EnsureProviderInConfig()
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
-
 	// perform API call
 	if flags.CmdVerbose {
 		fmt.Println(color.WhiteString("Fetching current cluster size"))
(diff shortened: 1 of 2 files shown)


commands/scale/cluster/command.go
```diff
@@ -174,16 +174,9 @@ func getClusterStatus(clusterID, activityName string) (*client.ClusterStatus, er
...
-		if flags.CmdVerbose {
-			fmt.Println(color.WhiteString("Fetching provider information"))
-		}
-
-		info, err := clientV2.GetInfo(auxParams)
-		if err != nil {
-			return nil, microerror.Mask(err)
-		}
-		config.Config.SetProvider(info.Payload.General.Provider)
+	err = clientV2.EnsureProviderInConfig()
```
xh3b4sd: When we do it this way we need to add it to all the commands. When we want to do it like this I would suggest a different entrypoint. Doing it for the "scale cluster" command in `getClusterStatus` is a snowflake and not straight forward. For other commands it would then look differently, which already smells. 

When we want to do it for every command separately, I would suggest a more straight forward entrypoint. But then it should always be the same. This would require a fair amount of refactoring I think. 

When you want to just make it work I thought we talked about the client initialization `NewWithConfig`. There we could then call `ensureProviderInConfig` before returning the client. It would then not be a so super cool design, but the most easy solution for now without a lot of refactoring.

marians: All right. Will change.

marians: Changed. PTAL

marians:
Seems I have to make tests work first...

marians:
Abandoning this here. It is not crucial currently, I can pick it up again later.

PR #380, opened 2019-07-23
Handle malformed response error in 'show cluster' command

marians:
This PR enables `gsctl show cluster` as implemented in the `support-v5` branch to work with the older API, not supporting v5 endpoints yet.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -484,3 +484,26 @@ selected_endpoint: ` + mockServer.URL
 		t.Errorf("Expected 'cluster-id', got %#v", clusterID)
 	}
 }
+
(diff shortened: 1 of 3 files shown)


marians:
@xh3b4sd Thanks! If you can point me to an example, I'd be interested.

xh3b4sd:
One example for testing HTTP handlers would be the response recorder concept in golang: https://golang.org/pkg/net/http/httptest/#ResponseRecorder. Another approach would be to inject a custom implementation of the transporter which then does not transport over network but only return what we want in a test: http://hassansin.github.io/Unit-Testing-http-client-in-Go#2-by-replacing-httptransport.

PR #381, opened 2019-07-24
In 'show nodepool', handle the situation where the instance type is unknown

marians:
The `show nodepool` command would panic in the case of an instance type that is not known in the `nodespec` package.

This PR adds handling for this situation as well as a test case.

The strategy is to still print as much information as possible, while showing that the instance type is not recognized.

### Preview

image

### Unrelated changes

To fix randomly failing CircleCI builds I updated the Golang version used in CIrcleCI and picked a bigger resource_class.


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.12.5
+      - image: circleci/golang:1.12.7
```
marians: Unrelated to the PR subject. I saw some unexplicable flaky behaviour in test runs and tried a remedy.

commands/show/nodepool/command_test.go
```diff
@@ -12,60 +13,97 @@ import (
...
-				}`))
-			default:
-				t.Errorf("Unsupported route %s called in mock server", r.URL.Path)
-				w.WriteHeader(http.StatusNotFound)
-				w.Write([]byte(`{"code": "RESOURCE_NOT_FOUND", "message": "Status for this cluster is not yet available."}`))
-			}
-		}
-	}))
-	defer mockServer.Close()
+	var testCases = []struct {
```
xh3b4sd: I see we have some evolution here. That is good. 👍

commands/show/nodepool/command.go
```diff
@@ -134,13 +134,13 @@ func fetchNodePool(args arguments) (*result, error) {
 		return nil, microerror.Mask(err)
 	}
 
-	res.instanceTypeDetails, err = awsInfo.GetInstanceTypeDetails(res.nodePool.NodeSpec.Aws.InstanceType)
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
+	var nodespecErr error
+	res.instanceTypeDetails, nodespecErr = awsInfo.GetInstanceTypeDetails(res.nodePool.NodeSpec.Aws.InstanceType)
```
xh3b4sd: Why can't we keep that simple and call it `err`?

commands/show/nodepool/command.go
```diff
@@ -134,13 +134,13 @@ func fetchNodePool(args arguments) (*result, error) {
...
 
-	res.sumCPUs = res.nodePool.Status.NodesReady * int64(res.instanceTypeDetails.CPUCores)
-	res.sumMemory = float64(res.nodePool.Status.NodesReady) * float64(res.instanceTypeDetails.MemorySizeGB)
+	if nodespecErr == nil {
```
xh3b4sd: We should differentiate between expected and unexpected errors to make the behaviour and experience for the user more predictable and reliable at the end of the day. With this there might be something going on that we miss in the first place only because the error handling is insufficient.

commands/show/nodepool/command.go
marians: Don't know. Got confused by a compiler message before. Changing.

commands/show/nodepool/command.go
marians: @xh3b4sd Please check whether https://github.com/giantswarm/gsctl/pull/381/commits/b40efd4409db86a76f5863c848f9e8a352c02f37 fixes this for you. Thanks!

commands/show/nodepool/command.go
```diff
@@ -136,12 +136,15 @@ func fetchNodePool(args arguments) (*result, error) {
 
 	res.instanceTypeDetails, err = awsInfo.GetInstanceTypeDetails(res.nodePool.NodeSpec.Aws.InstanceType)
 	if err != nil {
-		return nil, microerror.Mask(err)
+		// We deliberately ignore "instance type not found", but respect all other errors.
+		if !nodespec.IsInstanceTypeNotFoundErr(err) {
```
xh3b4sd: Let's please stick to https://github.com/giantswarm/fmt/blob/master/go/errors.md#matching-errors and have it without indentation. It is way easier to understand that way and the greater level of consistency makes it very easy to work with arbitrary projects.

marians: Done

marians:
Ready to re-review

marians:
Thanks!

PR #384, opened 2019-08-01
Add 'create nodepool' command

marians:

This PR adds the `gsctl create nodepool` command.

### Preview

```
Add a new node pool to a cluster.

This command allows to create a new node pool within a cluster. Node pools
are groups of wortker nodes sharing a common configuration. Create different
node pools to serve workloads with different resource requirements, different
availability zone spreading etc. Node pools are also scaled independently.

Note that some attributes of node pools cannot be changed later. These are:

- ID - will be generated on creation
- Availability zone assignment
- Instance type used

When creating a node pool, all arguments (except for the cluster ID) are
optional. Where an argument is not given, a default will be applied as
follows:

- Name: will be "Unnamed node pool <n>".
- Availability zones: the node pool will use 1 zone selected randomly.
- Instance type: the default instance type of the installation will be
  used. Check 'gsctl info' to find out what that is.
- Scaling settings: the minimum and maximum size will be set to 3,
  meaning that autoscaling is disabled.

Examples:

  The simplest invocation creates a node pool with default attributes
  in the cluster specified.

    gsctl create nodepool f01r4

  We recommend to set a descriptive name, to tell the node pool apart
  from others.

    gsctl create nodepool f01r4  --name "Batch jobs"

  Assigning the node pool to availabilty zones can be done in several
  ways. If you only want to ensure that several zones are used, specify
  a number liker like this:

    gsctl create nodepool f01r4 --num-availability-zones 2

  To set one or several specific zones to use, give a list of zone names
  or letters.

    gsctl create nodepool f01r4 --availability-zones b,c,d

  Here is how you specify the instance type to use:

    gsctl create nodepool f01r4 --aws-instance-type m4.2xlarge

  The initial node pool size is set by adjusting the lower and upper
  size limit like this:

    gsctl create nodepool f01r4 --nodes-min 3 --nodes-max 10

Usage:
  gsctl create nodepool <cluster-id> [flags]

Aliases:
  nodepool, np

Flags:
      --availability-zones strings   List of availability zones to use, instead of setting a number. Use comma to separate values.
      --aws-instance-type string     EC2 instance type to use for workers, e. g. 'm5.2xlarge'
  -h, --help                         help for nodepool
  -n, --name string                  name or purpose description of the node pool
      --nodes-max int                Maximum number of worker nodes for the node pool.
      --nodes-min int                Minimum number of worker nodes for the node pool.
      --num-availability-zones int   Number of availability zones to use. Default is 1.

Global Flags:
      --auth-token string   Authorization token to use
      --config-dir string   Configuration directory path to use (default "/Users/marian/.config/gsctl")
  -e, --endpoint string     The API endpoint to use
  -v, --verbose             Print more information
```

### DIFF
--- a/commands/create/nodepool/command.go
+++ b/commands/create/nodepool/command.go
@@ -2,7 +2,13 @@
 package nodepool
 
 import (
+	"fmt"
+	"os"
+
+	"github.com/fatih/color"
 	"github.com/giantswarm/gscliauth/config"
+	"github.com/giantswarm/gsclientgen/models"
+	"github.com/giantswarm/gsctl/client"
 	"github.com/giantswarm/gsctl/commands/errors"
 	"github.com/giantswarm/gsctl/flags"
 	"github.com/giantswarm/microerror"
@@ -84,7 +90,7 @@ Examples:
 
 	cmdAwsEc2InstanceType   string
 	cmdName                 string
-	cmdNumAvailabilityZones int
+	cmdAvailabilityZonesNum int
 	cmdAvailabilityZones    []string
 )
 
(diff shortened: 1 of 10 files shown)


commands/create/nodepool/command.go
```diff
@@ -0,0 +1,357 @@
...
+	"github.com/giantswarm/gsctl/commands/errors"
+	"github.com/giantswarm/gsctl/flags"
+	"github.com/giantswarm/microerror"
+	"github.com/spf13/cobra"
+)
+
+var (
+	// Command is the cobra command for 'gsctl create nodepool'
+	Command = &cobra.Command{
+		Hidden:  false, // TODO: set to false to make this command visible once the release is out.
```
xh3b4sd: It already is `false`.

commands/create/nodepool/command.go
```diff
@@ -0,0 +1,357 @@
...
+		Use:     "nodepool <cluster-id>",
+		Aliases: []string{"np"},
+		// Args: cobra.ExactArgs(1) guarantees that cobra will fail if no positional argument is given.
+		Args:  cobra.ExactArgs(1),
+		Short: "Create a node pool",
+		Long: `Add a new node pool to a cluster.
+
+This command allows to create a new node pool within a cluster. Node pools
+are groups of wortker nodes sharing a common configuration. Create different
```
xh3b4sd:
```suggestion
are groups of worker nodes sharing a common configuration. Create different
```

commands/create/nodepool/command.go
```diff
@@ -0,0 +1,357 @@
...
+	initFlags()
+}
+
+// initFlags initializes flags in a re-usable way, so we can call it from multiple tests.
+func initFlags() {
+	Command.ResetFlags()
+	Command.Flags().StringVarP(&cmdName, "name", "n", "", "name or purpose description of the node pool")
+	Command.Flags().IntVarP(&cmdAvailabilityZonesNum, "num-availability-zones", "", 0, "Number of availability zones to use. Default is 1.")
+	Command.Flags().StringSliceVarP(&cmdAvailabilityZones, "availability-zones", "", nil, "List of availability zones to use, instead of setting a number. Use comma to separate values.")
+	Command.Flags().StringVarP(&flags.CmdWorkerAwsEc2InstanceType, "aws-instance-type", "", "", "EC2 instance type to use for workers, e. g. 'm5.2xlarge'")
```
xh3b4sd: Why is there this `Cmd` prefix?

commands/create/nodepool/command.go
```diff
@@ -0,0 +1,357 @@
...
+  Assigning the node pool to availabilty zones can be done in several
+  ways. If you only want to ensure that several zones are used, specify
+  a number liker like this:
+
+    gsctl create nodepool f01r4 --num-availability-zones 2
+
+  To set one or several specific zones to use, give a list of zone names
+  or letters.
+
+    gsctl create nodepool f01r4 --availability-zones b,c,d
```
xh3b4sd: This is very AWS specific, isn't it? How would this look like with different providers? I am not sure this zone expanding feature is a good thing.

commands/create/nodepool/command.go
```diff
@@ -0,0 +1,357 @@
...
+	Command.Flags().StringVarP(&cmdName, "name", "n", "", "name or purpose description of the node pool")
+	Command.Flags().IntVarP(&cmdAvailabilityZonesNum, "num-availability-zones", "", 0, "Number of availability zones to use. Default is 1.")
+	Command.Flags().StringSliceVarP(&cmdAvailabilityZones, "availability-zones", "", nil, "List of availability zones to use, instead of setting a number. Use comma to separate values.")
+	Command.Flags().StringVarP(&flags.CmdWorkerAwsEc2InstanceType, "aws-instance-type", "", "", "EC2 instance type to use for workers, e. g. 'm5.2xlarge'")
+	Command.Flags().Int64VarP(&flags.CmdWorkersMin, "nodes-min", "", 0, "Minimum number of worker nodes for the node pool.")
+	Command.Flags().Int64VarP(&flags.CmdWorkersMax, "nodes-max", "", 0, "Maximum number of worker nodes for the node pool.")
+}
+
+// Arguments defines the arguments this command can take into consideration.
+type Arguments struct {
```
xh3b4sd: For the future, when we align command line tools more then this would look differently. In SIG Operator we try to align command line tools more to https://github.com/giantswarm/devctl.

commands/create/nodepool/command.go
```diff
@@ -0,0 +1,357 @@
...
+func printValidation(cmd *cobra.Command, positionalArgs []string) {
+	var err error
+	args, err := collectArguments(positionalArgs)
+	if err == nil {
+		err = verifyPreconditions(args)
+	}
+
+	if err == nil {
+		return
+	}
```
xh3b4sd: This here is extremely complicated to understand for me now. It would be nice to maybe talk about it and make it better. It also looks like that `printResults` does basically the same but implements the logic differently.

commands/create/nodepool/command.go
```diff
@@ -0,0 +1,357 @@
...
+
+	auxParams := clientWrapper.DefaultAuxiliaryParams()
+	auxParams.ActivityName = activityName
+
+	response, err := clientWrapper.CreateNodePool(args.ClusterID, requestBody, auxParams)
+
+	if err != nil {
+		// create specific error types for cases we care about
+		if clientErr, ok := err.(*clienterror.APIError); ok {
+			if clientErr.HTTPStatusCode == http.StatusForbidden {
```
xh3b4sd: Please see how we deal with errors in `fmt`. We can have custom error matchers for this.

commands/create/nodepool/command_test.go
```diff
@@ -0,0 +1,318 @@
...
+)
+
+// TODO:
+// cases which should succeed:
+//   - setting only scaling min or only max should result in proper setting
+//   - set AZ list to letters like "A,B,C"
+
+// TestCollectArgs tests whether collectArguments produces the expected results.
+func TestCollectArgs(t *testing.T) {
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
xh3b4sd: I would question why we need a mock server again. I think we should try to not write tests involving protocol layers if possible.

commands/create/nodepool/command_test.go
```diff
@@ -0,0 +1,318 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/spf13/afero"
+
+	"github.com/giantswarm/gsctl/commands/errors"
+	"github.com/giantswarm/gsctl/testutils"
+)
+
+// TODO:
```
xh3b4sd: When should this be addressed?

commands/create/nodepool/command.go
marians: Fixed

commands/create/nodepool/command.go
marians: Node pools is AWS only at first, so is the documentation. We'll tackle the added complexity when we get there.

We simplify the zone labels in Happa, too, both in node pool creation and display. We also simplify in node pools details display in `gsctl show nodepool`. IMO it is a big benefit in readability and will prevent errors. Allowing the same simplification for node pool creation to me is just consequential.

commands/create/nodepool/command.go
marians: From tradition, I would say, when  all flags were global variables. Could be changed in a separate PR.

commands/create/nodepool/command.go
marians: This is really how all gsctl commands are built currently.

`printValidation` is the pre-execution step which also triggers validation, then shows the result. If an error occured, it exits with a code 1.

`printResult` repeats the collectArguments step, but it does not have to expect and handle errors from that step, as those would already be handled before and it would not reach this stage.

commands/create/nodepool/command_test.go
marians: Now is a good time. Added this.

commands/create/nodepool/command_test.go
marians: It's there to satisfy a call to the API's info endpoint. Any idea how to prevent or mock this otherwise?

xh3b4sd: What do you want to test? We do not need to test TCP connections, TCP handshakes or dial or transport functions. This is all not relevant here I think. I think what we want to test is everything up to the API client call. When you want to verify behaviour based on client responses I also suggested fake transport implementations. Depends a bit on your goal. I am not saying what you do here is wrong and I very much appreciate that you have such a strong focus on automated verification of functionality. I just want to challenge the idea a little bit and see what we can do different.

xh3b4sd: For the record, this is not a blocking topic now for this PR.

commands/create/nodepool/command.go
marians: Added and applied error matchers to the `clienterror` package.

xh3b4sd: I see. Though, the error matcher interface is supposed to take `error`. Our `fmt` guideline is pretty clear about this. The type assertion should happen within the matcher, not outside of it. And then the error handling is nice and flat like everywhere else.

xh3b4sd: See the example for [`IsAlreadyTerminated`](https://github.com/giantswarm/aws-operator/blob/58efc26ee0f8d34b00527aa583a0c27cbd7996b3/service/controller/clusterapi/v29/resource/tccp/error.go#L23-L48) in the `aws-operator`.

marians: Thanks for the example! I moved the type assertion into the matchers.

marians:
Thanks for the input!

PR #386, opened 2019-08-01
Remove obsolete flag deprecation

marians:
The flag has been removed already some time ago, but this line was forgotten.


marians:
Merging here

Issue #160, opened 2017-11-29
Let output of `gsctl completion bash` be sourcable


marians:
To simplify loading of gsctl completion files for bash, we should support this:

    source <(gsctl completion bash)

So a subcommand `gsctl completion bash` should output completion code for bash to STDOUT.

This means we would probably have to remove the current function of `gsctl completion`.

We'll have to specify what `gsctl completion zsh` should do.

sjentzsch:
Bump.

Even the latest version won't work for me:

```
sjentzsch@SJ-Notebook:~$ source $(gsctl completion bash --stdout)
bash: #: No such file or directory
```

marians:
Thanks for the hint!

This doesn't make for a good first line in the output:

```
# bash completion for gsctl                                -*- shell-script -*-
```

marians:
@sjentzsch Can you try this please?

```
source <(gsctl completion bash --stdout)
```

sjentzsch:
@marians I see, yes, that one works. Thanks!

marians:
Done

PR #409, opened 2019-08-07
Fix syntax for bash completion in usage

marians:
For https://github.com/giantswarm/gsctl/issues/160

This fixes a usage info in the `gsctl completion` command.


marians:
Merging here

PR #410, opened 2019-08-08
Fix auth scheme via client initialization

marians:
This PR is a start to fix authentication issues introduced with recent refactorings.

The problem is that `github.com/giantswarm/gscliauth.config.ChooseScheme(endpoint, token)` always returns the scheme `giantswarm` when a non-empty token is passed via the function's parameters. The token parameter was meant to represent a token provided by the user via the command line, to override any config.

To better reflect, the command structure in this PR now distinguishes between one argument `token` and an additional one named `userProvidedToken`. The latter represents any token provided by the user as a command line argument. The existing `token` can be set based on authentication data from the config.

### TODO

- Apply the same fix to all other commands.

### DIFF
--- a/commands/info/command.go
+++ b/commands/info/command.go
@@ -33,10 +33,11 @@ var (
 
 // Arguments represents the arguments we can make use of in this command
 type Arguments struct {
-	scheme      string
-	token       string
-	verbose     bool
-	apiEndpoint string
+	apiEndpoint       string
+	scheme            string
+	token             string
+	userProvidedToken string
+	verbose           bool
 }
 
 // collectArguments returns an Arguments object populated by the user's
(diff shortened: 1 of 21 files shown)


commands/info/command.go
```diff
@@ -192,7 +194,7 @@ func info(args Arguments) (infoResult, error) {
 
 	// If an endpoint and a token is defined, we pull info from the API, too.
 	if args.apiEndpoint != "" && args.token != "" {
```
schjan: as this is the only place where `args.token` is used. What do you think about defining a boolean like `isUserProvided`?

As a different option we could also change the `client.NewWithConfig` function if that is easier than changing every single `gsctl` command.

marians: Thanks for the suggestion. I explored it and see it complicating things in many places. If important, let's please try that in another PR.

PR #411, opened 2019-08-09
Remove more v2 naming

marians:
We had still a lot of mentions of "v2" in tests and comments, which only made sense while we had two different versions of the API client in use in parallel. This PR should remove the remainders.


oponder (approved):
LGTM!

PR #412, opened 2019-08-09
Use clienterror matchers everywhere

marians:
To simplify and align error matching for errors generated in the `clienterrror` sub package, this PR 

- introduces some more matchers
- makes the matches handle more cases
- and applies them everywhere.

### DIFF
--- a/commands/create/keypair/command.go
+++ b/commands/create/keypair/command.go
@@ -3,7 +3,6 @@ package keypair
 
 import (
 	"fmt"
-	"net/http"
 	"os"
 	"regexp"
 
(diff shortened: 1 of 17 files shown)


marians:
FYI: Our tests are green. It's the coveralls API that currently fails with a 503.

commands/create/keypair/command.go
```diff
@@ -245,16 +244,14 @@ func createKeypair(args Arguments) (createKeypairResult, error) {
...
-				return result, microerror.Mask(errors.AccessForbiddenError)
-			} else if clientErr.HTTPStatusCode == http.StatusNotFound {
-				return result, microerror.Mask(errors.ClusterNotFoundError)
-			} else if clientErr.HTTPStatusCode == http.StatusForbidden {
-				return result, microerror.Mask(errors.AccessForbiddenError)
-			} else if clientErr.HTTPStatusCode == http.StatusBadRequest {
-				return result, microerror.Maskf(errors.BadRequestError, clientErr.ErrorDetails)
-			}
+		if clienterror.IsAccessForbiddenError(err) {
+			return result, microerror.Mask(errors.AccessForbiddenError)
```
xh3b4sd: It is an anti pattern to return an error and another value. By convention users of interfaces and code APIs should not expect or rely on any other return value in case an error is returned. By convention any other return value should be the zero value when an error is returned by some interface or code API.

xh3b4sd (approved):
This is certainly a nice improvement. Just left a note about some not so cool behaviour.

PR #413, opened 2019-08-12
Add 'delete nodepool' command

marians:

This PR adds the `gsctl delete nodepool` command.

### Usage preview

```
Delete new node pool from a cluster.

This command allows to delete a node pool.

Deleting a node pool means that all worker nodes in the pool will be drained,
cordoned and then terminated.

In case you are running workloads on the node pool you want to delete,
make sure that there is at least one other node pool with capacity to
schedule the workloads. Also check whether label selectors, taints and
tolerations will allow scheduling on other pool's worker nodes. The best
way to observe this is by manually cordoning and draining the pool's
worker nodes and checking workload's node assignments, before issuing
the 'delete nodepool' command.

Note: Data stored outside of persistent volumes will be lost and there is
no way to undo this.

Examples:

  To delete node pool 'np1id' from cluster 'f01r4', use this command:

    gsctl delete nodepool f01r4/np1id

  To prevent the confirmation questions, apply --force:

    gsctl delete nodepool f01r4/np1id --force

Usage:
  gsctl delete nodepool <cluster-id>/<nodepool-id> [flags]

Aliases:
  nodepool, np

Flags:
      --force   If set, no interactive confirmation will be required (risky!).
  -h, --help    help for nodepool
```

## Execution preview

```
$ gsctl delete nodepool m0ckd/abc
Do you really want to delete node pool 'abc' from cluster 'm0ckd'? [y/n]: y
Node pool 'abc' in cluster 'm0ckd' will be deleted as soon as all workloads are terminated.
```

### DIFF
--- a/client/clienterror/clienterror.go
+++ b/client/clienterror/clienterror.go
@@ -333,6 +333,31 @@ func New(err error) *APIError {
 		}
 	}
 
+	// delete node pool
+	if deleteNodePoolsUnauthorizedErr, ok := err.(*nodepools.DeleteNodePoolUnauthorized); ok {
+		return &APIError{
+			HTTPStatusCode: http.StatusUnauthorized,
+			OriginalError:  deleteNodePoolsUnauthorizedErr,
+			ErrorMessage:   "Unauthorized",
+			ErrorDetails:   "You don't have permission to delete this node pool.",
+		}
+	}
+	if deleteNodePoolsNotFoundErr, ok := err.(*nodepools.DeleteNodePoolNotFound); ok {
+		return &APIError{
+			HTTPStatusCode: http.StatusNotFound,
+			OriginalError:  deleteNodePoolsNotFoundErr,
+			ErrorMessage:   "Not found",
+			ErrorDetails:   "The cluster or node pool was not found or you don't have access to it.",
+		}
+	}
+	if deleteNodePoolsDefaultErr, ok := err.(*nodepools.DeleteNodePoolDefault); ok {
+		return &APIError{
+			HTTPStatusCode: deleteNodePoolsDefaultErr.Code(),
+			OriginalError:  deleteNodePoolsDefaultErr,
+			ErrorMessage:   deleteNodePoolsDefaultErr.Error(),
+		}
+	}
+
 	// create key pair
 	if createKeyPairUnauthorizedErr, ok := err.(*key_pairs.AddKeyPairUnauthorized); ok {
 		return &APIError{
--- a/commands/create/nodepool/command.go
+++ b/commands/create/nodepool/command.go
@@ -111,9 +111,9 @@ func initFlags() {
 	Command.Flags().StringVarP(&cmdName, "name", "n", "", "name or purpose description of the node pool")
 	Command.Flags().IntVarP(&cmdAvailabilityZonesNum, "num-availability-zones", "", 0, "Number of availability zones to use. Default is 1.")
 	Command.Flags().StringSliceVarP(&cmdAvailabilityZones, "availability-zones", "", nil, "List of availability zones to use, instead of setting a number. Use comma to separate values.")
-	Command.Flags().StringVarP(&flags.CmdWorkerAwsEc2InstanceType, "aws-instance-type", "", "", "EC2 instance type to use for workers, e. g. 'm5.2xlarge'")
-	Command.Flags().Int64VarP(&flags.CmdWorkersMin, "nodes-min", "", 0, "Minimum number of worker nodes for the node pool.")
-	Command.Flags().Int64VarP(&flags.CmdWorkersMax, "nodes-max", "", 0, "Maximum number of worker nodes for the node pool.")
+	Command.Flags().StringVarP(&flags.WorkerAwsEc2InstanceType, "aws-instance-type", "", "", "EC2 instance type to use for workers, e. g. 'm5.2xlarge'")
+	Command.Flags().Int64VarP(&flags.WorkersMin, "nodes-min", "", 0, "Minimum number of worker nodes for the node pool.")
+	Command.Flags().Int64VarP(&flags.WorkersMax, "nodes-max", "", 0, "Maximum number of worker nodes for the node pool.")
 }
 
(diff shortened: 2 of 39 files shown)


client/clienterror/clienterror.go
```diff
@@ -333,6 +333,31 @@ func New(err error) *APIError {
...
+			ErrorMessage:   "Unauthorized",
+			ErrorDetails:   "You don't have permission to delete this node pool.",
+		}
+	}
+	if deleteNodePoolsNotFoundErr, ok := err.(*nodepools.DeleteNodePoolNotFound); ok {
+		return &APIError{
+			HTTPStatusCode: http.StatusNotFound,
+			OriginalError:  deleteNodePoolsNotFoundErr,
+			ErrorMessage:   "Not found",
+			ErrorDetails:   "The cluster or node pool was not found or you don't have access to it.",
```
xh3b4sd: The other errors exclusively talk about node pools. This here mentions clusters as well. Is this how it should be?

commands/create/nodepool/command.go
```diff
@@ -9,12 +9,13 @@ import (
 	"github.com/fatih/color"
 	"github.com/giantswarm/gscliauth/config"
 	"github.com/giantswarm/gsclientgen/models"
+	"github.com/giantswarm/microerror"
+	"github.com/spf13/cobra"
```
xh3b4sd: This is very cool, but one of the instances in which we could have take a step back, and create a separate PR.

commands/delete/nodepool/command.go
```diff
@@ -0,0 +1,216 @@
...
+	clientWrapper, err := client.NewWithConfig(args.APIEndpoint, args.UserProvidedToken)
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	auxParams := clientWrapper.DefaultAuxiliaryParams()
+	auxParams.ActivityName = activityName
+
+	_, err = clientWrapper.DeleteNodePool(args.ClusterID, args.NodePoolID, auxParams)
+	if err != nil {
```
xh3b4sd:
```
if IsError(err) {
  //
} else if IsError(err) {
  // 
} else if err != nil {
  // 
}
```

commands/delete/nodepool/command.go
```diff
@@ -0,0 +1,216 @@
...
+		if subtext != "" {
+			fmt.Println(subtext)
+		}
+		os.Exit(1)
+	}
+
+	if deleted {
+		fmt.Println(color.GreenString("Node pool '%s' in cluster '%s' will be deleted as soon as all workloads are terminated.", args.NodePoolID, args.ClusterID))
+	} else if args.Verbose {
+		fmt.Println(color.WhiteString("Aborted."))
```
xh3b4sd: So we get the full stop here now. Why is that FMI? I have been told single word labels, titles, responses do not need that.

commands/delete/nodepool/command_test.go
```diff
@@ -0,0 +1,283 @@
...
+				APIEndpoint: "https://foo",
+				AuthToken:   "some-token",
+				ClusterID:   "clusterid",
+				NodePoolID:  "nodepoolid",
+				Force:       true,
+			},
+		},
+	}
+
+	yamlText := `last_version_check: 0001-01-01T00:00:00Z
```
xh3b4sd: Can we define that outside the function to make the test easier to read?

commands/delete/nodepool/command_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"testing"
+
+	"github.com/giantswarm/gsctl/commands/errors"
+	"github.com/giantswarm/gsctl/testutils"
+	"github.com/google/go-cmp/cmp"
+	"github.com/spf13/afero"
+)
+
+// TestCollectArgs tests whether collectArguments produces the expected results.
+func TestCollectArgs(t *testing.T) {
```
xh3b4sd:
```suggestion
func Test_collectArguments(t *testing.T) {
```

commands/delete/nodepool/command_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/giantswarm/gsctl/commands/errors"
+	"github.com/giantswarm/gsctl/testutils"
+	"github.com/google/go-cmp/cmp"
+	"github.com/spf13/afero"
+)
+
+// TestCollectArgs tests whether collectArguments produces the expected results.
```
xh3b4sd:
```suggestion
// Test_collectArguments tests whether collectArguments produces the expected results.
```

commands/delete/nodepool/command_test.go
```diff
@@ -0,0 +1,283 @@
...
+				t.Errorf("Case %d - Unexpected error '%s'", i, err)
+			}
+			if diff := cmp.Diff(tc.resultingArgs, args); diff != "" {
+				t.Errorf("Case %d - Resulting args unequal. (-expected +got):\n%s", i, diff)
+			}
+		})
+	}
+}
+
+// TestVerifyPreconditions tests cases where validating preconditions fails.
```
xh3b4sd:
```suggestion
// Test_verifyPreconditions tests cases where validating preconditions fails.
```

commands/delete/nodepool/command_test.go
```diff
@@ -0,0 +1,283 @@
...
+func TestVerifyPreconditions(t *testing.T) {
```
xh3b4sd:
```suggestion
func Test_verifyPreconditions(t *testing.T) {
```

client/clienterror/clienterror.go
marians: The thinking is that a 404 error could mean either that the node pool does not exist OR that the owning cluster does not exist. We don't know.

commands/delete/nodepool/command.go
marians: Probably you crept into my head and changed my thinking. Can't tell. Humans.

marians:
Done? PTAL @xh3b4sd

PR #415, opened 2019-08-12
Add 'update nodepool' command

marians:

This PR adds the `gsctl updatge nodepool` command to modify a node pool.

## Usage preview

```
$ gsctl update nodepool -h
Change the name or the scaling settings of a node pool.

Examples:

  gsctl update nodepool f01r4/75rh1 --name "General purpose M5"

  gsctl update nodepool f01r4/75rh1 --nodes-min 10 --nodes-max 20

Usage:
  gsctl update nodepool <cluster-id>/<nodepool-id> [flags]

Aliases:
  nodepool, np

Flags:
  -h, --help            help for nodepool
  -n, --name string     name or purpose description of the node pool
      --nodes-max int   Maximum number of worker nodes for the node pool.
      --nodes-min int   Minimum number of worker nodes for the node pool.

...
```

## Execution preview

```
$ gsctl update nodepool m0ckd/foo --nodes-min 1 --nodes-max 2 --name "my great pool"
Node pool 'my great pool' (ID 'foo') in cluster 'm0ckd' has been modified.
```

## TODO

- [ ] Add tests

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -473,6 +473,22 @@ func (w *Wrapper) GetNodePools(clusterID string, p *AuxiliaryParams) (*nodepools
 	return response, nil
 }
 
+// ModifyNodePool modifies a node pool.
+func (w *Wrapper) ModifyNodePool(clusterID, nodePoolID string, modifyNodePoolRequest *models.V5ModifyNodePoolRequest, p *AuxiliaryParams) (*nodepools.ModifyNodePoolOK, error) {
+	params := nodepools.NewModifyNodePoolParams().WithClusterID(clusterID).WithNodepoolID(nodePoolID).WithBody(modifyNodePoolRequest)
+	err := setParamsWithAuthorization(p, w, params)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	response, err := w.gsclient.Nodepools.ModifyNodePool(params, nil)
+	if err != nil {
+		return nil, clienterror.New(err)
+	}
+
+	return response, nil
+}
+
 // DeleteNodePool deletes a node pool.
 func (w *Wrapper) DeleteNodePool(clusterID, nodePoolID string, p *AuxiliaryParams) (*nodepools.DeleteNodePoolAccepted, error) {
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -473,6 +473,22 @@ func (w *Wrapper) GetNodePools(clusterID string, p *AuxiliaryParams) (*nodepools
...
+func (w *Wrapper) ModifyNodePool(clusterID, nodePoolID string, modifyNodePoolRequest *models.V5ModifyNodePoolRequest, p *AuxiliaryParams) (*nodepools.ModifyNodePoolOK, error) {
+	params := nodepools.NewModifyNodePoolParams().WithClusterID(clusterID).WithNodepoolID(nodePoolID).WithBody(modifyNodePoolRequest)
+	err := setParamsWithAuthorization(p, w, params)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	response, err := w.gsclient.Nodepools.ModifyNodePool(params, nil)
+	if err != nil {
+		return nil, clienterror.New(err)
```
xh3b4sd: I just noticed `clienterror.New(err)`. What is that? We should do the microerror wrapping here.

flags/flags.go
```diff
@@ -35,6 +35,9 @@ var (
 	// Full represents the switch to disable all output truncation, passed as a flag.
 	Full bool
 
+	// Name is the name of a cluster or node pool.
+	Name string
```
xh3b4sd: Why is it either or? Oh, and why are these flags even global? It would be nice to have the flags per command. So they are isolated. This is then also what the reference implementation in `devctl` does.

marians: I guess you figured it. It's because `flags.Name` is used as the name flag for creating a cluster in one command and the one for naming a node pool in another command.

It may have downsides, but it also has benefits. This way it's actually possible to link all places using, for example, the flag `--aws-instance-type` to a common place (`flags.WorkerAwsEc2InstanceType`).

client/client.go
marians: Let me put that on my refactoring list. This may affect many places, adapting which could make this PR huge.

marians:
@xh3b4sd Is this worth an approval from your side now?

xh3b4sd (approved):

PR #417, opened 2019-08-13
Rename and expose arguments in 'create cluster'

marians:
This PR adjusts the argument naming in the `create cluster` command to enable deep diffs in tests in a next step and for general alignment.


marians:
FYI going ahead here

PR #418, opened 2019-08-13
Add and refactor tests for 'create cluster'

marians:
This refactoring PR adds a specific test for the argument collection and validation functions and converts some other tests dealing with YAML parsing into one that into one that is table driven.

This adds a new dependencies that is needed to disable comparison of unexported struct attributes (here: `Arguments.fileSystem`).

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -78,3 +78,7 @@
   non-go = true
   go-tests = true
   unused-packages = true
+
+[[constraint]]
+  name = "github.com/google/go-cmp"
+  version = "0.3.0"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -23,32 +23,32 @@
 
 
 [[constraint]]
-  name = "github.com/Jeffail/gabs"
-  branch = "master"
-
-[[constraint]]
   name = "github.com/fatih/color"
   version = "1.1.0"
 
 [[constraint]]
-  branch = "master"
   name = "github.com/giantswarm/apiextensions"
+  branch = "master"
 
 [[constraint]]
   name = "github.com/giantswarm/columnize"
   branch = "master"
 
 [[constraint]]
-  branch = "master"
   name = "github.com/giantswarm/gscliauth"
+  branch = "master"
 
 [[constraint]]
-  branch = "master"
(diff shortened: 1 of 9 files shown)


Gopkg.toml
```diff
@@ -78,3 +78,7 @@
   non-go = true
   go-tests = true
   unused-packages = true
+
+[[constraint]]
+  name = "github.com/google/go-cmp"
```
xh3b4sd: It would make the life of everyone easier when the deps are sorted alphabetically. This is at least what we do in our other projects.

commands/create/cluster/command.go
```diff
@@ -27,12 +27,13 @@ import (
...
-	ClusterName              string
-	DryRun                   bool
-	FileSystem               afero.Fs
+	APIEndpoint       string
+	AuthToken         string
+	AvailabilityZones int
+	ClusterName       string
+	DryRun            bool
+	// fileSystem is not exposed to hide it from comparison in tests.
+	fileSystem               afero.Fs
```
xh3b4sd: I think you want to use https://godoc.org/github.com/google/go-cmp/cmp/cmpopts#IgnoreFields.

commands/create/cluster/command.go
```diff
@@ -154,6 +155,12 @@ Examples:
 )
 
 func init() {
+	initFlags()
```
xh3b4sd: Can you explain this initialization and the resetting below please?

marians: Flags initialization is in `initFlags` so it can be triggered from tests.

`Command.ResetFlags()` is called in `initFlags` in order to make sure that flags are in the same state whenever the flags are initialized. Without this, tests issuing `SetFlags` or `ParseFlags` would set flags which would remain persisted, so in tests we would have to execute `ResetFlags` explicitly. I find this the safer way.

commands/create/cluster/command.go
marians: check - thanks

Gopkg.toml
marians: Sorted.

marians:
@xh3b4sd PTAL

PR #421, opened 2019-08-15
'create cluster': Allow passing cluster definition to STDIN

marians:
Towards https://github.com/giantswarm/gsctl/pull/355

This enables passing cluster definition YAML to `gsctl create cluster` via the command line.

### Example

```
$ gsctl create cluster -f - <<EOF
owner: giantswarm
name: My test cluster
release: 7.1.1
EOF

$ gsctl create cluster -f - <myfile.yaml

$ cat myfile.yaml | gsctl create cluster -f -
```

Especially with the first method it becomes more handy to write more complex cluster creation examples which can be copy/pasted for re-use without the need to create a file, at least into bash.

### DIFF
--- a/commands/create/cluster/command.go
+++ b/commands/create/cluster/command.go
@@ -250,6 +250,13 @@ func printResult(cmd *cobra.Command, positionalArgs []string) {
 		case errors.IsNotEnoughWorkerNodesError(err):
 			headline = "Not enough worker nodes specified"
 			subtext = fmt.Sprintf("If you specify workers in your definition file, you'll have to specify at least %d worker nodes for a useful cluster.", limits.MinimumNumWorkers)
+		case errors.IsYAMLNotParseableError(err):
+			headline = "Could not parse YAML"
+			if aca.InputYAMLFile == "-" {
+				subtext = "The YAML data given via STDIN could not be parsed into a cluster definition."
+			} else {
+				subtext = fmt.Sprintf("The YAML data read from file '%s' could not be parsed into a cluster definition.", aca.InputYAMLFile)
+			}
 		case errors.IsYAMLFileNotReadableError(err):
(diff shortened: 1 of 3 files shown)


commands/create/cluster/command.go
```diff
@@ -250,9 +252,16 @@ func printResult(cmd *cobra.Command, positionalArgs []string) {
 		case errors.IsNotEnoughWorkerNodesError(err):
 			headline = "Not enough worker nodes specified"
 			subtext = fmt.Sprintf("If you specify workers in your definition file, you'll have to specify at least %d worker nodes for a useful cluster.", limits.MinimumNumWorkers)
+		case errors.IsYAMLNotParseableError(err):
```
xh3b4sd: FYI, by convention the matchers do not have the `Error` suffix. Check our `fmt` and all operator code.

commands/create/cluster/command.go
```diff
@@ -494,21 +505,40 @@ func createAddClusterBody(d *types.ClusterDefinition) *models.V4AddClusterReques
...
+			yamlString += scanner.Text() + "\n"
+		}
+
+		if err := scanner.Err(); err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		result.definition, err = readDefinitionFromYAML([]byte(yamlString))
+		if err != nil {
+			fmt.Printf("Error: %#v\n", err)
```
xh3b4sd: I think this is a leftover.

xh3b4sd (approved):

commands/create/cluster/command.go
marians: Some detail to pick up every day! :)

commands/create/cluster/command.go
marians: Yep, you're right. Thanks!

PR #422, opened 2019-08-16
Align build info in version and info commands

marians:
This fixes several problems in the `gsctl info` and `gsctl version` command output.

- Use the `buildinfo` package in the `gsctl version` command to actually output build info again
- Align the output between `gsctl info` and `gsctl version`

One goal is to hunch user to use `make` over `go build` so they have a git hash compiled in which can be helpful in a debugging situation.

### Preview

Output for a build done via `go build`:

image

Output for a build done via `make`:

image

### DIFF
--- a/buildinfo/buildinfo.go
+++ b/buildinfo/buildinfo.go
@@ -1,8 +1,13 @@
 // Package buildinfo defines variables which are set by the Go linker on build time using ldflags.
 package buildinfo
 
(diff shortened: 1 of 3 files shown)


xh3b4sd (approved):
I am one of the guys going with `go build` only FYI.

marians:
@xh3b4sd I won't hold it against you. :)

PR #424, opened 2019-08-29
Set authentication using auth writer.

oponder:
This changes the way we set the authentication header. The way we were doing it previously is not totally the 'correct' way of doing it.

We made authorization a required parameter, that's not really how the spec wants us to do it.

I made this change because it simplified the authentication state in Happa considerably, and it looks like it works ok with these changes in gsctl.

See: https://goswagger.io/generate/client.html#authentication

Related: https://github.com/giantswarm/gsclientgen/pull/47
Related: https://github.com/giantswarm/api-spec/pull/165


pipo02mix (approved):
LGTMIWFY

PR #425, opened 2019-09-03
Add API features info to capabilities

marians:
- Removes a hardcoded release version number from the capabilities package
- Adds a way to read the node pools minimum release version number from the info endpoint.


marians:
Re-ping @tuommaki

capabilities/capabilities.go
```diff
@@ -31,75 +27,15 @@ var (
...
-		},
 	}
 
 	// AllCapabilityDefinitions contains all the capabilities
-	AllCapabilityDefinitions = []CapabilityDefinition{
-		Autoscaling,
-		AvailabilityZones,
-		NodePools,
+	AllCapabilityDefinitions = []*CapabilityDefinition{
+		&Autoscaling,
```
tuommaki: These pointers bother me a bit because they practically expose these exported instances globally and therefore any code anywhere could modify them and accidentally break logic elsewhere.

I'm thinking that there should be probably non-pointer instances of these capabilities and when they are used together with the `Service` instance, then that service should keep its internal copy with completed information such as NodePools release versions etc.

capabilities/service.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/giantswarm/gsctl/client"
+)
+
+// Service provides methods to get more details on the installation's
+// and tenant cluster's capabilities.
+type Service struct {
+	// Provider identifies the installation's provide,r e. g. 'aws' or 'azure'.
+	Provider string
+
```
tuommaki: Here one should keep internal copy of initialized capabilities because it's this specific `Service` instance that initializes them and also that is used to get them and provide functionality to check whether specific compatibility exists.

capabilities/service.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	return nil
+}
+
+// GetCapabilities returns the list of capabilities that applies to a given release version,
+// copnsidering the installation's provider.
+// This is where some capability definitions get completed using information from the API.
+func (s *Service) GetCapabilities(releaseVersion string) ([]*CapabilityDefinition, error) {
```
tuommaki: I think here the code would be cleaner as well if returned slice of capabilities was not a pointer (as I don't think there's any reason for that to be needed).

capabilities/service.go
```diff
@@ -0,0 +1,101 @@
...
+			capabilities = append(capabilities, capability)
+		}
+	}
+
+	return capabilities, nil
+}
+
+// HasCapability returns true if the current context (provider, release) provides
+// the given capabililty.
+func (s *Service) HasCapability(releaseVersion string, capability *CapabilityDefinition) (bool, error) {
```
tuommaki:
```suggestion
func (s *Service) HasCapability(releaseVersion string, capability CapabilityDefinition) (bool, error) {
```

capabilities/errors.go
```diff
@@ -0,0 +1,39 @@
...
+var invalidConfigError = &microerror.Error{
+	Kind: "invalidConfigError",
+}
+
+// IsInvalidConfig asserts invalidConfigError.
+func IsInvalidConfig(err error) bool {
+	return microerror.Cause(err) == invalidConfigError
+}
+
+var couldNotFetchFeatures = &microerror.Error{
```
tuommaki: Maybe sort these alphabetically :slightly_smiling_face:

capabilities/service.go
```diff
@@ -0,0 +1,101 @@
...
+type Service struct {
+	// Provider identifies the installation's provide,r e. g. 'aws' or 'azure'.
+	Provider string
+
+	// Client is an API client the service can use to fetch info.
+	ClientWrapper *client.Wrapper
+}
+
+// New creates a new configured Service.
+func New(provider string, clientWrapper *client.Wrapper) (*Service, error) {
```
tuommaki: In general we always have `Config` struct which is then passed to `New()` function. I'm not personally super strict on this, but for general alignment that could make sense.

capabilities/service.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/gsctl/client"
+)
+
+// Service provides methods to get more details on the installation's
+// and tenant cluster's capabilities.
+type Service struct {
+	// Provider identifies the installation's provide,r e. g. 'aws' or 'azure'.
+	Provider string
```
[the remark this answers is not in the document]
marians: changed

capabilities/capabilities.go
marians: Changed this

capabilities/service.go
marians: done

capabilities/errors.go
marians: done

capabilities/service.go
marians: Sounds a bit much to me. I can see the benefit with more arguments though.

marians:
Please check whether this is what you expected.

capabilities/service.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// Service provides methods to get more details on the installation's
+// and tenant cluster's capabilities.
+type Service struct {
+	// provider identifies the installation's provide,r e. g. 'aws' or 'azure'.
+	provider string
+
+	// client is an API client the service can use to fetch info.
+	clientWrapper *client.Wrapper
```
tuommaki: As a last minor nitpick: This could maybe go up so these fields are also sorted alphabetically. :slightly_smiling_face:

marians: done

PR #429, opened 2019-09-04
'gsctl show cluster': Add handling for cluster not found error

marians:
This PR adds user-friendly error output in the case that the cluster could not be found via v5 nor v4  detail endpoints. Without this change, there is no output and gsctl exits with status 0 in this case.

## Preview

```
$ gsctl show cluster dfdg -v
Fetching details for cluster dfdg.
Fetching details for cluster via v5 API endpoint.
No usable v5 response. Fetching details for cluster via v4 API endpoint.
Cluster not found
Either there is no cluster with ID 'dfdg', or you have no access to it.
Please check whether the cluster is listed when executing 'gsctl list clusters'.
```

### DIFF
--- a/commands/show/cluster/command.go
+++ b/commands/show/cluster/command.go
@@ -345,7 +345,6 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
 	}
 
 	clusterDetailsV4, clusterDetailsV5, nodePools, clusterStatus, credentialDetails, err := getClusterDetails(args)
-	fmt.Printf("DEBUG: err = %#v\n", err)
 	if err != nil {
 		errors.HandleCommonErrors(err)
 
--- a/commands/show/cluster/command.go
+++ b/commands/show/cluster/command.go
@@ -345,8 +345,28 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
 	}
 
 	clusterDetailsV4, clusterDetailsV5, nodePools, clusterStatus, credentialDetails, err := getClusterDetails(args)
+	fmt.Printf("DEBUG: err = %#v\n", err)
 	if err != nil {
 		errors.HandleCommonErrors(err)
+
+		headline := ""
+		subtext := ""
+
+		switch {
+		case errors.IsClusterNotFoundError(err):
+			headline = "Cluster not found"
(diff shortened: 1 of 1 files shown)


commands/show/cluster/command.go
```diff
@@ -347,6 +347,25 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
...
 		errors.HandleCommonErrors(err)
+
+		headline := ""
+		subtext := ""
+
+		switch {
+		case errors.IsClusterNotFoundError(err):
+			headline = "Cluster not found"
+			subtext = fmt.Sprintf("Either there is no cluster with ID '%s', or you have no access to it.\n", args.clusterID)
+			subtext += "Please check whether the cluster is listed when executing 'gsctl list clusters'."
```
tuommaki: Why would `gsctl list clusters` help when cluster is not found?

commands/show/cluster/command.go
```diff
@@ -347,6 +347,25 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
...
 	if err != nil {
 		errors.HandleCommonErrors(err)
+
+		headline := ""
+		subtext := ""
+
+		switch {
+		case errors.IsClusterNotFoundError(err):
+			headline = "Cluster not found"
+			subtext = fmt.Sprintf("Either there is no cluster with ID '%s', or you have no access to it.\n", args.clusterID)
```
tuommaki:
```suggestion
			subtext = fmt.Sprintf("No cluster with ID '%s' exists.\n", args.clusterID)
```
While the security aspect here is user friendly, I think that in general it's not taken into account and when person doesn't have rights to some information, it mostly by intention.

commands/show/cluster/command.go
marians: For two reasons

1. If the cluster the user was trying to display is shown, the entry can be used to double-check the spelling of the ID
2. for double-checking that the cluster is in fact not there (not accessible or not existing)

commands/show/cluster/command.go
marians: We do it this way in most 404 cases, so I'd pass the suggestion in favour of consistency.

PR #433, opened 2019-09-16
Add v5 support (node pools) to the 'create cluster' command

marians:
Towards https://github.com/giantswarm/giantswarm/issues/5465

**Warning: This PR breaks the known user experience for cluster creation and requires release in a major version bump. Also it should not be merged before v5 (node pools) support lands in an active release.**

This PR introduces support for node pools to the `create cluster` command, plus additional changes:

- It is no longer possible to set worker node spec details via command line flags. Now a YAML definition has to be used.
- The `--dry-run` flag is removed without replacement.
- Worker node labels contained in the old (v4) cluster definition will no longer be passed to the creation request. They will be ignored. This has no effect on the result, as the API ignores them, too.
- Handling of YAML cluster definitions is now more strict. Unknown attributes will cause an error.
- Parsing of cluster definition YAML is implemented in a way that returns an interface which has to be casted into either a V4 or V5 specific definition.


commands/create/cluster/v4.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// perform API call
+	response, err := clientWrapper.CreateClusterV4(addClusterBody, auxParams)
+	if err != nil {
+		return "", "", microerror.Mask(err)
+	}
+
+	// success
+	location = response.Location
+	id = strings.Split(location, "/")[3]
```
tuommaki: Is this extracting cluster ID? Why not take it from the response? I feel that this is very fragile.

client/clienterror/clienterror.go
```diff
@@ -103,6 +103,32 @@ func New(err error) *APIError {
...
+		} else if ae.HTTPStatusCode == http.StatusBadRequest {
+			ae.ErrorMessage = "Invalid parameters"
+			ae.ErrorDetails = "The cluster cannot be created. Some parameter(s) are considered invalid.\n"
+			ae.ErrorDetails += "Details: " + createClusterDefaultErr.Payload.Message
+		}
+		return ae
+	}
+
+	// create cluster V5
+	if createClusterUnauthorizedErr, ok := err.(*clusters.AddClusterV5Unauthorized); ok {
```
tuommaki: I was thinking if we can somehow use the new `errors.As()` here to make this cleaner, but my quick test failed. Just dropping a note here in case someone figures directly how it would work. I'm ok with this for now.

commands/create/cluster/v4.go
marians: The V4 response unfortunately doesn't give us the ID in a better way. With V5 that's different.

tuommaki: Hmm? Is there maybe something that could be improved here then: https://github.com/giantswarm/api/blob/master/server/endpoint/v4/cluster/creator/endpoint.go#L144 ?

tuommaki: Oh, right, this is different from others and uses the `Encoder()` which omits that. Should we maybe add that to response body then?

marians: As this is v4 and we have been living with this for quite some time, I'd rather not touch this now.

tuommaki: Ok.

PR #436, opened 2019-09-27
Use Go modules

marians:
### TODO

- [x] Fix CI
- [x] Add info to development docs


marians:
@tuommaki Yes, thanks. The dep files should go away. Not needed any more, plus could cause confusion.

PR #437, opened 2019-09-30
Adapt 'gsctl create kubeconfig' to add support for v5 clusters

marians:
Towards https://github.com/giantswarm/giantswarm/issues/6447, https://github.com/giantswarm/giantswarm/issues/5465

This PR changes the behaviour if the `gsctl create kubeconfig` command

- to support v5 clusters, as it attempts to fetch v5 cluster details for the given cluster ID first, then only attempts v4 if that failed.
- some output regarding the crteated files is only printed when the `-v`/`--verbose` flag is applied, to make the output cleaner.

### Preview

image

### DIFF
--- a/commands/create/kubeconfig/command.go
+++ b/commands/create/kubeconfig/command.go
@@ -358,6 +358,40 @@ func createKubeconfigRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
 	}
 }
 
+// getClusterDetails fetches cluster details to get the tenant cluster API endpoint,
+// and attempts first v5 and then falls back to v4.
+func getClusterDetails(clientWrapper *client.Wrapper, clusterID string, auxParams *client.AuxiliaryParams, verbose bool) (string, error) {
+	// Try v5 first, then fall back to v4.
+	if verbose {
+		fmt.Println(color.WhiteString("Fetching cluster details using the v5 API endpoint"))
+	}
+	clusterDetailsResponseV5, err := clientWrapper.GetClusterV5(clusterID, auxParams)
+	if err != nil {
+		if clienterror.IsNotFoundError(err) {
+			// We ignore a 404 here and try v4 next.
+			if verbose {
+				fmt.Println(color.WhiteString("Cluster not found via the v5 endpoint. Attempting v4 endpoint."))
+			}
+			clusterDetailsResponseV4, err := clientWrapper.GetClusterV4(clusterID, auxParams)
+			if err != nil {
+				if clientErr, ok := err.(*clienterror.APIError); ok {
+					return "", microerror.Maskf(clientErr,
+						fmt.Sprintf("HTTP Status: %d, %s", clientErr.HTTPStatusCode, clientErr.ErrorMessage))
+				}
+
+				return "", microerror.Mask(err)
+			}
+
+			return clusterDetailsResponseV4.Payload.APIEndpoint, nil
+		}
+
+		// For all other errors than 404, we fail.
+		return "", microerror.Mask(err)
+	}
+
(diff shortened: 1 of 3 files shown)


commands/create/kubeconfig/command.go
```diff
@@ -367,28 +371,48 @@ func createKubeconfig(ctx context.Context, args Arguments) (createKubeconfigResu
...
 	}
 
-	var apiEndpoint string
-	{
-		if args.tenantInternal {
-			baseEndpoint := strings.Split(clusterDetailsResponse.Payload.APIEndpoint, urlDelimiter)[1:]
-			apiEndpoint = fmt.Sprintf("https://%s.%s", tenantInternalAPIPrefix, strings.Join(baseEndpoint, urlDelimiter))
-		} else {
-			apiEndpoint = clusterDetailsResponse.Payload.APIEndpoint
+	if result.apiEndpoint == "" {
```
tuommaki: Would it make sense to extract this to separate function and inline the call in if-case above? I find this logic a little difficult to follow, because first it works with V5, then if that fails it potentially just logs and next step is suddenly v4 call if some variable is unset so far.

I feel like these generated types require some wrapping so that the higher level business logic would be clean, but is a bit overkill at this point maybe.

marians: Changed the way cluster details are pulled.

commands/create/kubeconfig/command.go
```diff
@@ -354,6 +359,40 @@ func createKubeconfigRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
...
+		if clienterror.IsNotFoundError(err) {
+			// We ignore a 404 here and try v4 next.
+			if verbose {
+				fmt.Println(color.WhiteString("Cluster not found via the v5 endpoint. Attempting v4 endpoint."))
+			}
+			clusterDetailsResponseV4, err := clientWrapper.GetClusterV4(clusterID, auxParams)
+			if err != nil {
+				if clientErr, ok := err.(*clienterror.APIError); ok {
+					return "", microerror.Maskf(clientErr,
+						fmt.Sprintf("HTTP Status: %d, %s", clientErr.HTTPStatusCode, clientErr.ErrorMessage))
```
tuommaki:
```suggestion
						"HTTP Status: %d, %s", clientErr.HTTPStatusCode, clientErr.ErrorMessage)
```
I probably missed this earlier, but with `Maskf()` `fmt.Sprintf()` is not needed. `f` in `Maskf` means formatting :)

commands/create/kubeconfig/command.go
```diff
@@ -354,6 +359,40 @@ func createKubeconfigRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
...
 }
 
+// getClusterDetails fetches cluster details to get the tenant cluster API endpoint,
+// and attempts first v5 and then falls back to v4.
+func getClusterDetails(clientWrapper *client.Wrapper, clusterID string, auxParams *client.AuxiliaryParams, verbose bool) (string, error) {
+	// Try v5 first, then fall back to v4.
+	if verbose {
+		fmt.Println(color.WhiteString("Fetching cluster details using the v5 API endpoint"))
+	}
+	clusterDetailsResponseV5, err := clientWrapper.GetClusterV5(clusterID, auxParams)
```
tuommaki: With the latest version of this function, this could be clearer if there's:
```go
if err == nil {
    return clusterDetailsResponseV5.Payload.APIEndpoint, nil
}
```

So then you can remove one level indentation and focus on V4 rest of the function.

marians: Changed

PR #439, opened 2019-10-07
App client support

ferrandinand:
Creates support methods for CRUD apps.


pipo02mix (approved):
LGTM, did not you were working on it, nice. Probably the hardest part here is to come later with a good command naming and structure

marians:
@ferrandinand I don't see any specific errors added in `clienterror`. Please keep that in mind for following PRs using these functions.

ferrandinand:
I merged but I will try to add them as it seems a reasonable concern

PR #442, opened 2019-10-10
Re-arranging the scale cluster command

marians:
This PR aligns the `scale cluster` command implementation with other commands.

Plus it removes the Autoscaling feature detection, as the autoscaling API (scaling.min, scaling.max) is now supported by all tenant clusters.

Changes include:

- Use the command's `PreRun` hook for validation, calling a `printValidation` function, which calls a ` verifyPreconditions` function
- Have minimal parameters for the `collectArguments` function
- Removal of all deprecated/unsupported flags
  - `--memory-gb`
  - `--num-cpus`
  - `--storage-gb`
- Removes setting of config.Config.Provider, as this is no longer needed. It was used for Autoscaling feature detection.
- Fixes https://github.com/giantswarm/giantswarm/issues/7137


commands/scale/cluster/command.go
```diff
@@ -54,296 +52,255 @@ Examples:
...
-	}
-	status, err := clientWrapper.GetClusterStatus(clusterID, auxParams)
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
+	errors.HandleCommonErrors(err)
 
-	return status, nil
-}
+	headline := ""
```
tuommaki:
```suggestion
	var headline string
```
Extremely minor thing, but something that I noticed thinking. Given our usual practice to have blocks like following:
```go
var foo TypeBar
{
    f, err := findOutFoo()
    if err != nil {
        return microerror.Mask(err)
    }
    foo = f
}
```

It's somehow easier to anticipate following logic when the variable is just declared instead of introduced and explicitly initialized when in here the actually value assignation is done in following block. No strong opinions here but something that I found myself thinking about.

commands/scale/cluster/command.go
```diff
@@ -54,296 +52,255 @@ Examples:
...
+	subtext := ""
```
tuommaki:
```suggestion
	var subtext string
```

commands/scale/cluster/command.go
```diff
@@ -352,53 +309,15 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
...
-	}
-
 	// Actually make the scaling request to the API.
-	details, err := scaleCluster(args)
+	result, err := scaleCluster(args)
 	if err != nil {
 		errors.HandleCommonErrors(err)
 		client.HandleErrors(err)
 
+		headline := ""
```
tuommaki:
```suggestion
		var headline string
```

commands/scale/cluster/command.go
```diff
@@ -352,53 +309,15 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
...
+		subtext := ""
```
tuommaki:
```suggestion
		var subtext string
```

commands/scale/cluster/command.go
```diff
@@ -352,53 +309,15 @@ func printResult(cmd *cobra.Command, cmdLineArgs []string) {
...
+
```
tuommaki:
```suggestion
```

commands/scale/cluster/command.go
```diff
@@ -54,296 +52,255 @@ Examples:
...
-	status, err := clientWrapper.GetClusterStatus(clusterID, auxParams)
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
+	errors.HandleCommonErrors(err)
 
-	return status, nil
-}
+	headline := ""
+	subtext := ""
```
tuommaki:
```suggestion
```
I'd maybe _zap_ empty line between variable declaration and value assignation block.

PR #444, opened 2019-10-11
improve error handling by returning zero values

xh3b4sd:


marians:
Can you provide some context regarding what problem this is solving?

kopiczko (approved):

xh3b4sd:
A general best practise is to return the zero value of any return value in case a non nil error is returned. APIs should follow this always and no code should ever expect any other value in case an error is returned. Not complying with this makes code potentially more fragile, confusing and dangerous.

PR #461, opened 2019-11-05
Handle error 500 consistently in all commands

marians:
This PR aligns the handling of Internal Server Errors (HTTP 500) in all commands.

- In the `clienterror` package we make sure the error "message" part is passed to the returned APIError object.
- In the `client.HandleErrors()` function we make sure to handle the internal error
- We apply `client.HandleErrors()` consistently in every command.

### Before

```
gsctl scale cluster ...
[GET /v4/clusters/{cluster_id}/][500] getCluster default  &{Code:INTERNAL_ERROR Message:Something bad happened which we'll explain somewhere else.}
```

### After

```
gsctl scale cluster ...
An internal error occurred.
Something bad happened which we'll explain somewhere else.
```

### DIFF
--- a/commands/upgrade/cluster/command.go
+++ b/commands/upgrade/cluster/command.go
@@ -113,6 +113,7 @@ func upgradeClusterValidationOutput(cmd *cobra.Command, cmdLineArgs []string) {
 	err := validateUpgradeClusterPreconditions(args, cmdLineArgs)
 
 	if err != nil {
+		client.HandleErrors(err)
 		errors.HandleCommonErrors(err)
 
 		switch {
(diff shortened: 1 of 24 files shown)


marians:
Re-ping @giantswarm/sig-front-end

commands/upgrade/cluster/command.go
```diff
@@ -113,6 +113,7 @@ func upgradeClusterValidationOutput(cmd *cobra.Command, cmdLineArgs []string) {
 	err := validateUpgradeClusterPreconditions(args, cmdLineArgs)
 
 	if err != nil {
+		client.HandleErrors(err)
```
oponder: Perhaps out of scope, but all these commands have to learn this correct order of handling errors. Can we extract that to a single call which then encapsulates this knowledge, so that in the future, if something changes, you can make an edit in a single place, instead of having to go through all commands again?

marians: Yeah. It started very modular as there are (very few) commands which don't instantiate the `client` package. Might be an option for further optimization.

PR #463, opened 2019-11-12
Make harder the Cluster deletion

pipo02mix:


pipo02mix:
I have checked the effort to output the name and it will be a big deal as we don't have a centralized place to get cluster details (for both versions). I would like to make another step later, but for now, go with it 

cc/ @marians

confirm/confirm.go
```diff
@@ -33,3 +33,28 @@ func Ask(s string) bool {
...
+// then press enter.
+func AskStrict(s string, c string) bool {
+	reader := bufio.NewReader(os.Stdin)
+
+	fmt.Printf("%s: ", color.YellowString(s))
+
+	for {
+		response, err := reader.ReadString('\n')
+		if err != nil {
+			log.Fatal(err)
```
fiunchinho: I think the error [should be masked](https://github.com/giantswarm/fmt/blob/master/go/errors.md)

confirm/confirm.go
```diff
@@ -33,3 +33,28 @@ func Ask(s string) bool {
...
 }
+
+// AskStrict asks the user for confirmation. A user must type the expected confirmation text and
+// then press enter.
+func AskStrict(s string, c string) bool {
+	reader := bufio.NewReader(os.Stdin)
+
+	fmt.Printf("%s: ", color.YellowString(s))
+
+	for {
```
fiunchinho: I think it would work the same way without the for loop, and removing it would mean having less code!

fiunchinho (approved):

confirm/confirm.go
```diff
@@ -33,3 +35,28 @@ func Ask(s string) bool {
...
+		}
+
+		switch strings.ToLower(strings.TrimSpace(response)) {
+		case c:
+			return true
+		case "n", "no":
+			return false
+		default:
+			fmt.Printf(color.YellowString("The input entered does not match."))
+			fmt.Printf(color.YellowString("Try again or abort the deletion typing 'n[o]': "))
```
oponder:
```suggestion
			fmt.Printf(color.YellowString("Try again or abort by typing 'n' or 'no': "))
```

oponder (approved):
LGTM but didn't test. Have a problem in my go env that I don't want to rabbit hole to fix right nw.

PR #464, opened 2019-11-12
Fix cluster definition in 'create cluster' usage

marians:
The usage example shows an attribute `release`, which should i fact be named `release_version`.


oponder (approved):
👍

PR #466, opened 2019-11-13
Provide specific error for 'list nodepools' when cluster is v4

marians:
For https://github.com/giantswarm/giantswarm/issues/6447

Previously, using `gsctl list nodepools` on a v4 cluster would produce this not very helpful output:

```
$ gsctl list np ry4mo
Not found
```

With this change, we check whether a v4 cluster with the given ID exists. If yes, we show this:

```
This cluster does not support node pools.
Node pools cannot be listed for this cluster. Please use 'gsctl show cluster' to get information on worker nodes.
```


ograu (approved):
LGTM!

PR #468, opened 2019-11-18
Show deleted clusters with the 'gsctl list clusters' command

marians:
Towards https://github.com/giantswarm/giantswarm/issues/7516

The goal of that story is to raise awareness for clusters still using resources after deletion has been initialised, until all resources are actually removed.

This adds a `--show-deleting` flag to the command and a conditional `DELETING SINCE` column to the cluster listing, showing the deletion date if available. Deleted clusters are also marked red when deleted more than 24 hours ago.

Note: When using `-o json` for JSON output, deleted clusters will be listed. The `--show-deleted` flag is not applicable in that case.

### Preview

image

image


xh3b4sd:
Just some thoughts to consider. 

* `docker ps -a` shows dead containers. Do we want to show deleted clusters using a flag or always? 
* Initially I thought there should maybe two tables be shown to have a more clear visual separation. Uncertain if this is a good practise. 
* Should there be some kind of sorting? Should deleted clusters be shown first, or last? The list of Cloud Formations in the AWS Console does not change sorting because of the stack state.

marians:
I got some more feedback in SIG UX and we decided to ask a broader audience.

Here are two versions presented:

image

- (A) would show deleted clusters in the output always, as implemented in this PR. The red highlighting should only occur after at least 24 hours from the delete_date, to differentiate it from a cluster that has been deleted a couple of minutes ago and to indicate "something isn't right".

- (B) Would hide deleted clusters by default, but would show an additional info line at the bottom of the output. Users would have to add `--show-deleted` to list all clusters, including the deleted ones. The result would look like (A).

(B) is currently the winner, so I'll adapt the PR accordingly.

marians:
Regarding sorting, this table is sorted by ID. That has not changed in this PR.

marians:
Adapted the behaviour to what we discussed / what the majority voted for. Please take another look.

marians:
Just found another detail to fix

commands/list/clusters/command.go
```diff
@@ -156,40 +175,104 @@ func getClustersOutput(args Arguments) (string, error) {
...
+				color.RedString(cluster.ID),
+				color.RedString(cluster.Owner),
+				color.RedString(cluster.Name),
+				color.RedString(releaseVersion),
+				color.RedString(created),
+			}
+			if args.showDeleted {
+				fields = append(fields, color.RedString(deleted))
+			}
+		} else {
```
pipo02mix: I would extract repeated logic before the condition but no strong feelings

pipo02mix (approved):
LGTM in general, minor comment and also the flag name does not reflect really the purposed (deleted vs deleting) but if none else has concerns, go ahead. Plz wait for sig operator approval

commands/list/clusters/command.go
marians: Good point! Changed.

marians:
@pipo02mix If I get you right, you are saying that "deleted" suggests something different than what the user is getting here, correct? Fair point.

marians:
Modified after @pipo02mix 's comment:

image

PR #469, opened 2019-11-19
Fix usage info in 'create cluster' command

marians:
This fixes the false statement that by default the cluster would have no node pools.

### DIFF
--- a/commands/create/cluster/command.go
+++ b/commands/create/cluster/command.go
@@ -127,12 +127,12 @@ general defaults:
 - Release: the latest release is used.
 - Workers
   - On AWS and when using the latest release, and when not specifying node pool
-	details via a clsuter definition, the cluster will be created with a 
-	default node pool. You may define node pools in your cluster definition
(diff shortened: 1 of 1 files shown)


commands/create/cluster/command.go
```diff
@@ -126,9 +126,10 @@ general defaults:
 
 - Release: the latest release is used.
 - Workers
-  - On AWS and when using the latest release, the cluster will be created
-    without any node pools. You may define node pools in your cluster
-    definition YAML or add node pools one by one using 'gsctl create nodepool'.
+  - On AWS and when using the latest release, and when not specifying node pool
+	details via a clsuter definition, the cluster will be created with a
```
xh3b4sd:
```suggestion
	details via a cluster definition, the cluster will be created with a 
```

xh3b4sd: Is the indentation on purpose? Looks inconsistent. 🙈

PR #476, opened 2019-12-04
Picking up improvements from master for v0.17.x

marians:
Here we are collecting changed made to master, to be released in a release 0.17.5.

The reason we still keep 0.17.x is that node pools is not generally released yet and master contains breaking changes to the `gsctl create cluster` command which only make real sense together with node pools.


marians:
Re-ping @giantswarm/sig-front-end - Anybody able to give this a BS check?

PR #478, opened 2019-12-11
Improve handling of "malformed response" error

marians:
Towards https://github.com/giantswarm/giantswarm/issues/4430

This PR attempts to handle the `Malformed response` error consistently in all commands (with the exception of `gsctl ping` as that one is using a different client.

The error's cause is usually that some Ingress controller responds with an HTML page showing some "forbidden" message. The cause here is often that a VPN connection is not set up and the client is connecting from a network that has no permission to access the API.

Before:

image

After:

image

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -503,7 +503,7 @@ func TestMalformedResponse(t *testing.T) {
 	}
 
 	_, err = gsClient.GetClusterV4("cluster-id", nil)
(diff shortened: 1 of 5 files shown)


ubergesundheit (approved):
LGTM

oponder (approved):
LGTM!

marians:
Thanks! 🍪

PR #481, opened 2019-12-11
Show web UI URL in 'gsctl show cluster' output

marians:
Towards https://github.com/giantswarm/giantswarm/issues/6058

This adds displaying the web UI URL for cluster details in the `gsctl show cluster` output.

### Preview

image

### DIFF
--- a/commands/show/cluster/command.go
+++ b/commands/show/cluster/command.go
@@ -21,6 +21,7 @@ import (
 	"github.com/giantswarm/gsctl/flags"
 	"github.com/giantswarm/gsctl/nodespec"
 	"github.com/giantswarm/gsctl/util"
+	"github.com/giantswarm/gsctl/webui"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


commands/show/cluster/command.go
```diff
@@ -482,6 +492,10 @@ func printV5Result(args Arguments, details *models.V5ClusterDetailsResponse,
 		clusterTable = append(clusterTable, formatCredentialDetails(credentialDetails)...)
 	}
 
+	if webUIURL != "" {
```
oponder: I wouldn't have expected to need these bits of code twice in the same file / command.

Can it be extracted into a function `webUIURL()` ?

webui/webui.go
```diff
@@ -0,0 +1,60 @@
...
+
+	hostNameParts := strings.Split(parsed.Hostname(), ".")
+
+	// replace first part with 'happa' if it is 'api'
+	if len(hostNameParts) < 2 {
+		return "", microerror.Mask(unsupportedHostNameError)
+	}
+
+	if hostNameParts[0] != apiStandardHostName {
+		return "", microerror.Maskf(unsupportedHostNameError, "host name must start with 'api'")
```
oponder: I guess if you take the trouble to make `apiStandardHostName` then you
```suggestion
		return "", microerror.Maskf(unsupportedHostNameError, fmt.Sprintf("host name must start with '%s'", apiStandardHostName))
```

oponder (approved):
Cool! just some suggestions you might want to fix before merging.

commands/show/cluster/command.go
marians: You mean `webUIURL, _ := webui.ClusterDetailsURL(args.apiEndpoint, details.ID, details.Owner)` ? That's called two times, but with different arguments.

PR #484, opened 2019-12-14
Update k8s dependencies to 1.16.4

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/7675


Makefile
```diff
@@ -24,7 +24,7 @@ TESTBIN := build/bin/${BIN}-${GOOS}-${GOARCH}
 all: build
 
 prebuild:
-	go get -u github.com/gobuffalo/packr/packr
+	go install github.com/gobuffalo/packr/packr
```
sslavic: It was very annoying to pull latest master, run `make clean install` and be left with modified go.mod and go.sum files, signaling non-reproducible build. Part of mystery was that CI for gsctl already ensures `go mod tidy` is run and fails the build if either `go.mod` or `go.sum` were untidy. Running `go mod tidy` locally would clean things up again, so the difference in versions specified and versions used in the build wasn't due to difference in `go` version used in CI build and locally.

Analysis uncovered root cause to be `-u` switch in `go get` command, it would update packr dependency and its transitive dependencies, so not exactly respecting packr version specified in `go.mod` + `go.mod` files.

To fix that issue and to make it more obvious that `packr` binary is supposed to be installed, `go get -u` is replaced with `go install`.

Also, `tools.go` was added to track `packr` tool dependency explicitly using canonical Go modules approach (see [related official docs](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module)).

Makefile
```diff
@@ -24,7 +24,7 @@ TESTBIN := build/bin/${BIN}-${GOOS}-${GOARCH}
...
 	packr
```
sslavic: This one was a mystery why is it being run after being installed as there was no obvious side-effect in the project own files/directories; thought maybe it's just used to verify `packr` tool was installed correctly, that it can be run, but there was no other place in `Makefile` or in CircleCI build definition where it's being used.

It turned out this `packr` run without any additional parameters/commands is actually a must, part of [using packr "the hard way"](https://github.com/gobuffalo/packr#building-a-binary-the-hard-way), which as a result generates `-packr.go` files for all `packr` `Box`es for a project, all of which for `gsctl` (and `opsctl`) are defined in the shared `gscliauth` library, and since `gsctl` dependencies are managed with Go modules, `-packr.go` files actually end up in go modules cache, under `$GOPATH/pkg/mod` (or `$HOME/go/pkg/mod` if `GOPATH` is not set), so outside of the project directory. Not sure would it make sense to add a comment above or on the same line with `packr` binary usage with reference to ["the hard way"](https://github.com/gobuffalo/packr#building-a-binary-the-hard-way), maybe it's obvious to others, until now `packr` wasn't in my toolbox.

rossf7 (approved):
LGTM

marians:
Thanks @sslavic !

sslavic:
You're welcome! I didn't make a release out of it yet, so it's not available to most. @marians when does it make sense to release this change?

marians:
What benefits does this bring to end users? My understanding so far was that it is an internal change.

sslavic:
Yes, internal change, adds support for our future k8s 1.16 based releases once they become available.

marians:
Then no release is needed. We'll release 0.18.0 anyway in the next few days.

PR #495, opened 2020-01-15
Check for provider not supported when scaling

MarcelMue:
We noticed while testing, that the API returns `not supported` for g8s versions greater than 10 - this breaks the behaviour here for providers other than AWS.

related slack thread: https://gigantic.slack.com/archives/CHAHY6VMF/p1579081042011500

### DIFF
--- a/commands/scale/cluster/command.go
+++ b/commands/scale/cluster/command.go
@@ -171,7 +171,7 @@ func verifyPreconditions(args Arguments) error {
 		}
 
 		_, err = clientWrapper.GetClusterV5(args.ClusterID, auxParams)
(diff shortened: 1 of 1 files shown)


QuentinBisson (approved):
LGTM

tfussell (approved):

tfussell:
Do we need to consider changing the API or is this sufficient?

MarcelMue:
> Do we need to consider changing the API or is this sufficient?

I think this should work for now - I don't really want to touch the API as tooling should just use the V4 endpoint for all providers except AWS nodepools.

marians (approved):

PR #496, opened 2020-01-16
Actually check for bad request instead of not supported

MarcelMue:
Apparently I was testing the wrong thing manually. It actually needs to match against a bad request error to work. 

Example here: https://github.com/giantswarm/gsctl/blob/11c2ebd40ad6c38a8fe0932ef4be2e206a58f879/commands/create/kubeconfig/command.go#L375

### DIFF
--- a/commands/scale/cluster/command.go
+++ b/commands/scale/cluster/command.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/giantswarm/gsctl/client"
+	"github.com/giantswarm/gsctl/client/clienterror"
 	"github.com/giantswarm/gsctl/commands/errors"
 	"github.com/giantswarm/gsctl/confirm"
 	"github.com/giantswarm/gsctl/flags"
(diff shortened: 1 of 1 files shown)


tfussell:
Where is this error returned in the API? Just to be sure that they match.

MarcelMue:
Here is the code that defines it as a `400` return value:
https://github.com/giantswarm/api/blob/a4e9ef3f6c433559abd227d4eaabb219b804f08a/server/server.go#L431-L434

I do not understand which error lib I should be using in gsctl - but the one used in the kubeconfig command actually matches against `400`. 
Why there are different libs used is not clear to me.

I tested this with this branch now  and was able to scale to 4 workers:
```
gsctl scale cluster a44c3  --num-workers 4                 
The cluster is being scaled
The cluster limits have been changed from min=3 and max=3 to min=4 and max=4 workers.
```

```
gsctl show cluster a44c3
ID:                        a44c3
Name:                      testing-marcel-conformance
Created:                   2020 Jan 14, 15:24 UTC
Organization:              giantswarm
Kubernetes API endpoint:   https://api.a44c3.k8s.ghost.westeurope.azure.gigantic.io
Release version:           11.0.0
Availability Zones:        1
Worker VM size:            Standard_D4s_v3
Worker node scaling:       pinned at 4
Worker nodes running:      5
CPU cores in workers:      20
RAM in worker nodes (GB):  85.90
Web UI:                    https://happa.g8s.ghost.westeurope.azure.gigantic.io/organizations/giantswarm/clusters/a44c3
```

tfussell:
Looks like `github.com/giantswarm/gsctl/commands/errors` is for returning errors from `gsctl` command execution while `github.com/giantswarm/gsctl/client/clienterror` is for errors from the API so I think this is right.

PR #499, opened 2020-01-27
Display a warning if the endpoint uses the HTTP protocol

axbarsan:
Resolve https://github.com/giantswarm/giantswarm/issues/8502

- Display a warning if the client is using an endpoint with a HTTP protocol
- Store arguments at package level for commands that require pre-run validation

Some warnings have been added to tests, since we're using a HTTP endpoint for testing, to make tests pass


oponder (commented):
LGTM! Might be good to do some exploratory testing manually as well though.

axbarsan:
Tested manually and everything's working as expected 👍

oponder (approved):
👍

PR #500, opened 2020-01-28
Create test for complete v4 cluster creation

axbarsan:
Reference https://github.com/giantswarm/giantswarm/issues/8601

Add test for complete v4 cluster creation, to check if flags like `AvailabilityZones` get passed through


oponder (approved):
LGTM!

PR #501, opened 2020-02-03
Implement 'delete endpoint' command

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/8442

### DIFF
--- a/commands/delete/command.go
+++ b/commands/delete/command.go
@@ -13,7 +13,7 @@ var (
 	Command = &cobra.Command{
 		Use:   "delete",
 		Short: "Delete things",
-		Long:  `Lets you delete a cluster, a node pool or an API endpoint`,
+		Long:  `Lets you delete a cluster, a node pool, or an API endpoint`,
 	}
 )
 
--- a/commands/delete/command.go
+++ b/commands/delete/command.go
@@ -4,6 +4,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/giantswarm/gsctl/commands/delete/cluster"
+	"github.com/giantswarm/gsctl/commands/delete/endpoint"
 	"github.com/giantswarm/gsctl/commands/delete/nodepool"
 )
 
@@ -12,11 +13,12 @@ var (
 	Command = &cobra.Command{
 		Use:   "delete",
 		Short: "Delete things",
-		Long:  `Lets you delete a cluster or node pool`,
+		Long:  `Lets you delete a cluster, a node pool or an API endpoint`,
 	}
 )
 
 func init() {
 	Command.AddCommand(cluster.Command)
 	Command.AddCommand(nodepool.Command)
+	Command.AddCommand(endpoint.Command)
 }
--- a/commands/delete/endpoint/command.go
+++ b/commands/delete/endpoint/command.go
@@ -55,11 +55,14 @@ Example:
 		PreRun: printValidation,
 		Run:    printResult,
 	}
-
-	arguments Arguments
 )
 
 func init() {
+	initFlags()
+}
+
+func initFlags() {
+	Command.ResetFlags()
 	Command.Flags().BoolVarP(&flags.Force, "force", "", false, "If set, no interactive confirmation will be required (risky!).")
 }
 
@@ -67,7 +70,7 @@ func init() {
 // If errors occur, error info is printed to STDOUT/STDERR
 // and the program will exit with non-zero exit codes.
 func printValidation(cmd *cobra.Command, args []string) {
-	arguments = collectArguments(args)
+	arguments := collectArguments(args)
 
 	err := validatePreconditions(arguments)
 	if err != nil {
@@ -108,8 +111,10 @@ func validatePreconditions(args Arguments) error {
 	return nil
 }
 
-// interprets arguments/flags, eventually submits delete request
+// interprets arguments/flags, deletes endpoint
 func printResult(cmd *cobra.Command, args []string) {
+	arguments := collectArguments(args)
+
 	deleted, err := deleteEndpoint(arguments)
 	if err != nil {
 		client.HandleErrors(err)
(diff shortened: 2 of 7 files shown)


commands/delete/command.go
```diff
@@ -12,11 +13,12 @@ var (
 	Command = &cobra.Command{
 		Use:   "delete",
 		Short: "Delete things",
-		Long:  `Lets you delete a cluster or node pool`,
+		Long:  `Lets you delete a cluster, a node pool or an API endpoint`,
```
oponder:
```suggestion
		Long:  `Lets you delete a cluster, a node pool, or an API endpoint`,
```

Up in these woods we use the oxford comma :P

🇬🇧

commands/delete/endpoint/command_test.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+			if diff := cmp.Diff(tc.resultingArgs, args); diff != "" {
+				t.Errorf("Case %d - Resulting args unequal. (-expected +got):\n%s", i, diff)
+			}
+		})
+	}
+}
+
+// TestDeleteEndpointSuccess runs test case that are supposed to succeed
+func TestDeleteEndpointSuccess(t *testing.T) {
```
oponder: Should we do a test of printResult as well? Just noticing the decrease in coverage.

oponder (changes requested):
Thanks! Just have a grammar suggestion and a testing question.

commands/delete/endpoint/command_test.go
axbarsan: Yes, that's on its way 😬

oponder (approved):
LGTM IIWFY!

marians:
I am wondering whether the confirmation is really needed.

**PRO**: We do it with all other deletions, so users may expect it.

**CON**: Deleting an endpoint removes the entry and auth token from the config. It is not destroying any actual data in the primary sense, like deleting a cluster may.

But when in doubt, it's probably fine to have the confirmation.

axbarsan:
As you said, we have confirmations in all of our deletions, so the user will expect it to be here, also.
Deleting it accidentally would not make you lose valuable data, but will make you lose time, which is equally important 👍

commands/delete/endpoint/command.go
```diff
@@ -0,0 +1,175 @@
...
+		case errors.IsEndpointMissingError(err):
+			headline = "No API endpoint specified"
+			subtext = "See --help for usage details."
+		case errors.IsCouldNotDeleteEndpointError(err):
+			headline = "The API endpoint could not be deleted."
+			subtext = "You might try again in a few moments. If that doesn't work, please contact the Giant Swarm support team."
+			subtext += "Sorry for the inconvenience!"
+		default:
+			headline = err.Error()
+		}
```
marians: Here in `printValidation` we only have to handle the errors that can occur in `validatePreconditions`.

`errors.IsCouldNotDeleteEndpointError` is not returned by `validatePreconditions` so should not be handled here. Instead it has to be handled in the `printResult` function. Also `client.HandleErrors(err)` isn't needed (but also doesn't hurt) here, as there is no client usage in `validatePreconditions`.

marians (changes requested):

marians:
Confirmation is fine with me, however I think having to type some string is over-cautious.

```
./gsctl delete endpoint gauss
Do you really want to delete API endpoint 'gauss'? Please type the endpoint name to confirm:
```

A simple n/y (case insensitive) should do here.

```
./gsctl delete endpoint gauss
Do you really want to delete API endpoint 'gauss'? [y/N]:
```

"N" is capitalized to indicate that it is assumed to be the default when nothing is entered.

axbarsan:
> Confirmation is fine with me, however I think having to type some string is over-cautious.
> 
> ```
> ./gsctl delete endpoint gauss
> Do you really want to delete API endpoint 'gauss'? Please type the endpoint name to confirm:
> ```
> 
> A simple n/y (case insensitive) should do here.
> 
> ```
> ./gsctl delete endpoint gauss
> Do you really want to delete API endpoint 'gauss'? [y/N]:
> ```
> 
> "N" is capitalized to indicate that it is assumed to be the default when nothing is entered.

I've added this functionality to all the `yes/no` confirms

marians:
Wording nitpick:

    The API endpoint '<endpoint-or-alias>' deleted successfully.

could be changed to

    The API endpoint '<endpoint-or-alias>' has been deleted.

Found a little UX problem we might want to work around in this PR.

When deleting the currently selected endpoint, and then issuing commands like `gsctl list clusters`, this error occurs:

```
You are not logged in.
Use 'gsctl login' to login or '--auth-token' to pass a valid auth token.
```

The error message is misleading and I created https://github.com/giantswarm/giantswarm/issues/8843 for that.

However, we could do more here in the context of the `delete endpoint` command to prevent users from running into this. When deleting the current endpoint, I suggest to show this additional output:

```
The API endpoint '<endpoint-or-alias>' has been deleted.

Note: You have no API endpoint selected now. Use 'gsctl select endpoint' to select one.
```

confirm/confirm.go
```diff
@@ -20,17 +20,19 @@ func Ask(s string) bool {
...
 
 		response, err := reader.ReadString('\n')
 		if err != nil {
 			log.Fatal(microerror.Mask(err))
 		}
 
 		switch strings.ToLower(strings.TrimSpace(response)) {
-		case "y", "yes":
+		case "y", "yes", "Y", "YES", "Yes":
 			return true
```
marians: As you convert the string to be tested to lower case, there should be node need to test `"Y", "YES", "Yes"` , right?

marians (approved):

confirm/confirm.go
axbarsan: You're right, I didn't catch that. Thank you 👍

PR #502, opened 2020-02-05
Add tab completion to select endpoint command

### LINKED ISSUE #159: Tab completion should expand endpoint URLs
Almost all commands support setting the endpoint to use with the `--endpoint` or `-e` flag. In addition, the `gsctl select endpoint <endpoint>` command expects the endpoint as a positional argument.

To greatly simplify entry, the endpoint string should be Tab-completable.

axbarsan:
Fixes #159 

This introduces bash auto-completion for the `select endpoint` command, and the `--endpoint` flag. Pressing `TAB` will un-fold the list of existing endpoints.

Unfortunately, this doesn't work in `zsh` due to `cobra` limitations.

**Preview**

image

image

### DIFF
--- a/util/completion_test.go
+++ b/util/completion_test.go
@@ -16,9 +16,9 @@ var getCompletionFnTests = []struct {
 	output string
 }{
 	{
-		"test",
-		"ls",
-		`
+		fName: "test",
+		fBody: "ls",
+		output: `
 test() {
 	ls
 }`,
@@ -41,17 +41,17 @@ var setFlagCompletionFnTests = []struct {
 	output string
 }{
 	{
-		"test",
-		"ls",
-		`
+		fName: "test",
+		fBody: "ls",
+		output: `
 test() {
 	ls
 }`,
 	},
 	{
-		"test2",
-		"ls",
-		`
+		fName: "test2",
+		fBody: "ls",
+		output: `
 test() {
 	ls
 }
@@ -60,9 +60,9 @@ test2() {
 }`,
 	},
 	{
-		"test",
-		"ls",
-		`
+		fName: "test",
+		fBody: "ls",
+		output: `
 test() {
 	ls
 }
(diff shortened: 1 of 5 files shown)


util/completion_test.go
```diff
@@ -0,0 +1,122 @@
...
+	Use: "test-command",
+}
+
+var getCompletionFnTests = []struct {
+	fName  string
+	fBody  string
+	output string
+}{
+	{
+		"test",
```
oponder: maybe we should use the `key: "value"` form here to make things a bit more readable..

like 
```
{
fName: "test",
...
}
```

oponder (approved):
LGTM, but did not test

axbarsan:
Ping @marians 

Added a description, let me know what you think

marians:
@axbarsan Big ❤️ for working on this and even providing a test. Unfortunately I cannot make it work in my local setup. Perhaps it would be best to pair on this.

Another note: https://github.com/giantswarm/gsctl/issues/159 is about completion of the endpoint in all comands. This PR is only about one command. So please keep #159 open.

axbarsan:
> @axbarsan Big ❤️ for working on this and even providing a test. Unfortunately I cannot make it work in my local setup. Perhaps it would be best to pair on this.
> 
> Another note: #159 is about completion of the endpoint in all comands. This PR is only about one command. So please keep #159 open.

Hi Marian, this actually provides auto-completion for both `select endpoint` and any command that uses the `--endpoint` flag, so I believe it should close that issue.

As with getting it to work, this is how I've tested it:

**gsctl source folder**
```bash
go build
./gsctl completion bash --stdout > ./.gsctl-completion
source .gsctl-completion
```

Also, be sure to have `bash_completion` working
```bash
source $(brew --prefix)/etc/bash_completion
```

After this, everything _should_ work

If it does not, it might be worth deleting your existing completion configuration (if you have it somewhere) from the system.

marians:
Even more awesome! Working for me, too. I was missing `source $(brew --prefix)/etc/bash_completion`. Thank you!

marians (approved):
Great!

PR #507, opened 2020-02-25
Fix passing of email argument to login command

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/9085

This PR fixes a regression introduced with https://github.com/giantswarm/gsctl/pull/499 that would prevent the email argument from being passed to the API call.

### DIFF
--- a/commands/login/command.go
+++ b/commands/login/command.go
@@ -67,8 +67,12 @@ type Arguments struct {
 	verbose     bool
 }
 
-func collectArguments() Arguments {
+func collectArguments(positionalArgs []string) Arguments {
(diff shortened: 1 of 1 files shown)


commands/login/command.go
```diff
@@ -186,7 +187,8 @@ func login(loginArgs Arguments) (loginResult, error) {
 
 // loginRunOutput executes the login logic and
 // prints output and sets the exit code.
-func loginRunOutput(cmd *cobra.Command, args []string) {
+func loginRunOutput(cmd *cobra.Command, positionalArgs []string) {
+	arguments := collectArguments(positionalArgs)
```
marians: Adding line 191 is the main fix

PR #508, opened 2020-02-27
Display an error message if there's no API endpoint selected

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/8843

Displays an error if there is no API endpoint selected while trying to execute a command that needs a connection to the API endpoint

**Preview**

image


marians (approved):
What I see looks good, thank you! Harder to judge: is anything missing?

axbarsan:
> What I see looks good, thank you! Harder to judge: is anything missing?

I would take the optimistic route and say that nothing is missing 😬

Issue #188, opened 2018-01-11
Accept cluster name as argument, as an alternative to a cluster ID


marians:
Some users seem to expect to be able to use cluster names in `gsctl` commands. E. g.

    gsctl scale cluster "My cluster name"

Currently, cluster names are not required to be unique, so there _might_ be cases where the name would be ambigious. That case could be handled with a specific error message.

To resolve a cluster name to its ID, we would require an additional call to `GET /v4/clusters/`.

marians:
If we can do this in half a day, let's do it.

axbarsan:
Commands that need this implemented

- [ ] Create keypair
- [ ] Create kubeconfig
- [ ] Create nodepool
- [ ] Delete cluster
- [ ] Delete nodepool
- [ ] List keypairs
- [ ] List nodepools
- [ ] Scale cluster
- [ ] Show cluster
- [ ] Show nodepool
- [ ] Update cluster
- [ ] Update nodepool
- [ ] Upgrade cluster

axbarsan:
All the PRs have been merged. Testing each command again, on `master`, before closing this

axbarsan:
Tested again, every command is working like before 👍

PR #509, opened 2020-02-27
Add support for using cluster name for the delete cluster command

axbarsan:
Towards https://github.com/giantswarm/gsctl/issues/188

At the moment, this only provides support for the delete cluster command
- Renamed `ClusterIDMissing` errors to `ClusterNameOrIDMissing`
- Before executing the delete request, it gets all the clusters and checks to see if there is a cluster with the provided ID or name in there
- Changed error messages to also say cluster **name**, not just **ID**

When a cluster is deleted, the confirmation will state the name/ID provided by the user. So if the user provided an ID, the confirmation will show the ID; or a name, if he provided a name.

**Preview**
* Deleting a cluster with the name
image

* Deleting a cluster with the ID
image

* Trying to delete a cluster with an ID/name that doesn't exist
image

* Trying to delete a cluster that has the same name as another existing one
image

* Saving cluster IDs in the cache file (`~/.config/gsctl/clustercache.yaml`)
image

As a side note, this only includes the `delete cluster` command, because I want to know that my implementation is right, before copying and pasting it to the other commands.
Also, I would like to do 1-2 commands per PR, to keep them small, so we could prevent any unwanted bugs.

### DIFF
--- a/clustercache/clustercache.go
+++ b/clustercache/clustercache.go
@@ -22,31 +22,31 @@ const (
 	listClustersActivityName = "list-clusters"
 	clusterCacheFileName     = "clustercache.yaml"
 
-	cacheDuration = time.Hour * 24 * 7 // 7 days
+	cacheDuration = time.Hour * 24 * 7 // 7 days.
 	timeLayout    = time.RFC3339
 )
 
 // EndpointCache stores the IDs stored in an
-// endpoint-specific cache, and also its expiry date
+// endpoint-specific cache, and also its expiry date.
 type EndpointCache struct {
 	Expiry string   `yaml:"expiry"`
 	IDs    []string `yaml:"ids"`
 }
 
 // Endpoints stores a map with the keys being API endpoints,
-// and the values being caches
+// and the values being caches.
 type Endpoints map[string]EndpointCache
 
-// Cache is the file structure of the cluster cache file
+// Cache is the file structure of the cluster cache file.
 type Cache struct {
 	Endpoints Endpoints `yaml:"endpoints"`
 }
 
 // GetID gets the cluster ID for a provided name/ID
-// by checking in both the user cache and on the API
+// by checking in both the user cache and on the API.
 func GetID(endpoint string, clusterNameOrID string, clientWrapper *client.Wrapper) (string, error) {
 	// Check if the cluster ID is already in the cache,
-	// and skip the API request if it is
+	// and skip the API request if it is.
 	isInCache := IsInCache(endpoint, clusterNameOrID)
 	if isInCache {
 		return clusterNameOrID, nil
@@ -55,7 +55,7 @@ func GetID(endpoint string, clusterNameOrID string, clientWrapper *client.Wrappe
 	auxParams := clientWrapper.DefaultAuxiliaryParams()
 	auxParams.ActivityName = listClustersActivityName
 
-	// Get a list of all clusters
+	// Get a list of all clusters.
 	response, err := clientWrapper.GetClusters(auxParams)
 	if err != nil {
 		switch {
@@ -71,14 +71,14 @@ func GetID(endpoint string, clusterNameOrID string, clientWrapper *client.Wrappe
 	}
 
 	var (
-		// IDs that correspond to the same cluster name
+		// IDs that correspond to the same cluster name.
 		matchingIDs []string
-		// All IDs returned from the response
+		// All IDs returned from the response.
 		allClusterIDs = make([]string, 0, len(response.Payload))
 	)
 	for _, cluster := range response.Payload {
 		allClusterIDs = append(allClusterIDs, cluster.ID)
-		// Check if this is the cluster we're looking for
+		// Check if this is the cluster we're looking for.
 		if matchesValidation(clusterNameOrID, cluster) {
 			matchingIDs = append(matchingIDs, cluster.ID)
 		}
@@ -90,8 +90,8 @@ func GetID(endpoint string, clusterNameOrID string, clientWrapper *client.Wrappe
 		// There are no IDs that correspond to that cluster name
 		return "", microerror.Mask(errors.ClusterNotFoundError)
 	} else if len(matchingIDs) > 1 {
-		// There are multiple IDs that correspond to that cluster name
-		// Help the user decide which cluster to pick
+		// There are multiple IDs that correspond to that cluster name.
+		// Help the user decide which one to pick.
 		_, id := handleNameCollision(clusterNameOrID, response.Payload)
 
 		return id, nil
@@ -100,7 +100,7 @@ func GetID(endpoint string, clusterNameOrID string, clientWrapper *client.Wrappe
 	return matchingIDs[0], nil
 }
 
-// New creates a new Cache object
+// New creates a new Cache object.
 func New() *Cache {
 	c := &Cache{}
 	c.Endpoints = Endpoints{}
@@ -109,7 +109,7 @@ func New() *Cache {
 }
 
 // IsInCache checks if a cluster ID is present in the
-// persistent cluster cache
+// persistent cluster cache.
 func IsInCache(endpoint string, ID string) bool {
 	var (
 		err                error
@@ -117,7 +117,7 @@ func IsInCache(endpoint string, ID string) bool {
 
 		now = time.Now()
 	)
-	// Read the existing cache file
+	// Read the existing cache file.
 	existing, err := read(config.FileSystem)
 	if err != nil {
 		return false
@@ -125,12 +125,12 @@ func IsInCache(endpoint string, ID string) bool {
 
 	c := existing.Endpoints[endpoint]
 	for _, cID := range c.IDs {
-		// Check if the cache is expired
+		// Check if the cache is expired.
 		endpointExpiration, err = time.Parse(timeLayout, c.Expiry)
 		if err != nil || now.After(endpointExpiration) {
 			return false
 		}
-		// This is the one we're looking for
+		// This is the one we're looking for.
 		if cID == ID {
 			return true
 		}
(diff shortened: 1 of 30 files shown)


ubergesundheit (approved):
Cool!

commands/delete/cluster/command.go
```diff
@@ -185,32 +186,63 @@ func printResult(cmd *cobra.Command, args []string) {
...
+		}
+		if clienterror.IsAccessForbiddenError(err) {
+			return "", microerror.Mask(errors.AccessForbiddenError)
+		}
+
+		return "", microerror.Mask(err)
+	}
+
+	for _, cluster := range response.Payload {
+		if cluster.ID == clusterNameOrID || cluster.Name == clusterNameOrID {
```
oponder: Here we basically go for the first found match right?

I think since names do not have to be unique, we should fail with a "Cluster name is ambiguous" type of failure message when the name you chose is present more than once.

It would be pretty rare, but I can imagine someone naming their clusters "Dev" "Production" and "Staging" and then having multiple organisations with those clusters all with the same name or something like that. So it's not too far fetched.

axbarsan: Ahhh yesss, I actually had this in my notes but completely forgot about this

Thank you for noticing, I'll update and throw an error if there are multiple clusters with the same names

oponder (changes requested):
See my comment on the name ambiguity edge case. Might need a bit of UX consideration in how we present that message. 

Like the coolest would be to offer some kind of picker that is like 
```
Found more than one cluster called "My Cluster", please type the id of the cluster that you would like to delete:

CLUSTER ID    ORGANIZATION    NAME
123ab         testing         My Cluster 
456cd         production      My Cluster 
```

But if that needs more UX consideration, then for now just failing and asking the user to use a cluster ID instead would be ok with me.

axbarsan:
> See my comment on the name ambiguity edge case. Might need a bit of UX consideration in how we present that message.
> 
> Like the coolest would be to offer some kind of picker that is like
> 
> ```
> Found more than one cluster called "My Cluster", please type the id of the cluster that you would like to delete:
> 
> CLUSTER ID    ORGANIZATION    NAME
> 123ab         testing         My Cluster 
> 456cd         production      My Cluster 
> ```
> 
> But if that needs more UX consideration, then for now just failing and asking the user to use a cluster ID instead would be ok with me.

I like that, I'll try to implement it

commands/delete/cluster/command.go
marians: Yep, even if rare, it would be fatal, so I consider that absolutely necessary.

marians:
So my understanding is we basically have to list clusters in every command that needs a cluster ID argument currently, before we can issue the API call. Because we don't know in an input like

    gsctl create kubeconfig -c a1b2c

whether `a1b2c` is a cluster ID or cluster name.

In installations with many organizations and clusters we see this taking several seconds (in one case, more than 5). So every command would get delayed by that duration. To me, that is too much of a downside.

What are other ways we can think of? Some ideas:

- Maintain a cache of last seen cluster IDs: If input `a1b2c` is a hit in our cache, we assume that it's an ID and don't try to resolve name to ID first. (We could re-build that cache periodically and/or whenever we call `GET /v4/clusters/`)
- Allow users to turn off the feature? When turned off, the behaviour should be just like it is currently.

axbarsan:
Yes, we do fetch all clusters in every command.

I believe caching wouldn't be an ideal solution, especially if we encounter situations like this:
- Creating a cluster in `gsctl`
- Deleting it through `happa` or direct API call and creating a new one with the same name
- Deleting it in `gsct` would throw an error, because in the cache there's another ID stored for that name, and the API will error out, saying it does not exist

> Allow users to turn off the feature? When turned off, the behaviour should be just like it is currently.
I would vote for this

Also, would it make sense to have a `-n|--name` param that will specify the name? So it will only make an extra API call if you do it that way

marians:
The caching idea was not about caching cluster names. Rather only cache cluster IDs we really got from the API. So every hit on that cache could be considered a valid cluster ID that at least _has_ existed at some point.

axbarsan:
Oh I get it now, so we'll cache the cluster IDs, so if we provided an ID, it would check the cache and if it's there, it would not make the API call. That sounds good, I can start on that

axbarsan:
I added support for name collisions, which also prints a table with the colliding clusters, and then you can enter the ID of the cluster that you want to use.

Updated the description with the preview.

Next: Adding the ID cache

axbarsan:
Also added caching and unit tests for caching

marians:
After I opened the  `util` package, I learned that it isn't good Go practice to have such a catch-all package. In Go, devs expect explanatory package names. Look at all the folders on the top level. It would fit the current layout of the repo to just add a package like `cache` or, more descriptive, `cluster-cache`.

util/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+
+const (
+	listClustersActivityName = "list-clusters"
+	clusterCacheFileName     = "clustercache"
+)
+
+// GetClusterID gets the cluster ID for a provided name/ID
+// by checking in both the user cache and on the API
+func GetClusterID(clusterNameOrID string, clientWrapper *client.Wrapper) (string, error) {
+	isID := IsInClusterCache(clusterNameOrID)
```
marians: "isID" implies that we can find out whether this is an ID. In fact, we can only find out if it is an ID we know. So `isKnownID` or `isCachedID` would fit better.

util/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+		confirmed, id := handleNameCollision(clusterNameOrID, response.Payload)
+		if !confirmed {
+			return "", nil
+		}
+
+		return id, nil
+
+	default:
+		return clusterIDs[0], nil
+	}
```
marians: I think an if / else if construct and a simple return after that would make this easier to read, no?

util/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+// cluster IDs, for commands that take both cluster names and IDs
+func CacheClusterIDs(c ...string) {
+	fs := config.FileSystem
+
+	existingC := make(chan []string)
+	go func() {
+		e, _ := readClusterCache(fs)
+		existingC <- e
+	}()
+	existing := <-existingC
```
marians: Why use a channel to read a file's content?

marians: And why read the existing cache content at all? When you cache the last cached items, why do you care what was in the cache before? This will grow the list forever, no?

util/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func readClusterCache(fs afero.Fs) ([]string, error) {
+	filePath := path.Join(config.ConfigDirPath, clusterCacheFileName)
+	output, err := afero.ReadFile(fs, filePath)
+	if err != nil {
+		return []string{}, err
+	}
+
+	c := strings.Split(string(output), ",")
```
marians: The separator should be a constant in the package. BTW, I would prefer a newline. That would make the file much more handy if one should ever have to debug something in a text editor.

util/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+
+	return c, nil
+}
+
+func writeClusterCache(fs afero.Fs, c ...string) error {
+	filePath := path.Join(config.ConfigDirPath, clusterCacheFileName)
+	output := []byte(strings.Join(c, ","))
```
marians: As above, please use a separator constant

marians:
I haven't looked at everything yet, so might have missed it. One more questions regarding the cluster cache: Does the cache ever expire? Would cluster IDs get re-fetched when the cache is older than a certain duration?

One thing to keep in mind is that the cache is not endpoint/installation specific. So what is a valid ID in one installation might not be one in another. But with cluster IDs as random as they are, I don't see that as a problem.

util/cluster.go
axbarsan: My initial thinking was that we might call this method in different places, not just when fetching all clusters. But now that you pointed it out, it probably won't get used in other places, so it doesn't really make sense to append to the existing cache (which in the first case, it should've had a limit)

axbarsan: And about the channels, I wanted to somehow make cache writing non-blocking to the actual commands that use it, but I think I'm missing something in the actual implementation, because it doesn't to work like that

axbarsan:
Thank you for the thorough review, @marians . I really appreciate the suggestions you've made.

I didn't take into consideration expiry dates and endpoint-specific IDs, but I think it may be worth switching to a YAML structure to introduce those, to make it adding other functionality easier to do in the future

axbarsan:
* Made the cache save to a YAML file
* Added different IDs per API endpoint
* Added expiration dates for cluster caches

Updated the description with a preview of the file

marians:
Nitpick: Some people in the company will be much happier to find sentences in comments ending in a full stop. Do as you wish.

clustercache/clustercache.go
```diff
@@ -0,0 +1,240 @@
...
+
+	CacheIDs(endpoint, allClusterIDs)
+
+	if matchingIDs == nil {
+		// There are no IDs that correspond to that cluster name
+		return "", microerror.Mask(errors.ClusterNotFoundError)
+	} else if len(matchingIDs) > 1 {
+		// There are multiple IDs that correspond to that cluster name
+		// Help the user decide which cluster to pick
+		_, id := handleNameCollision(clusterNameOrID, response.Payload)
```
marians: If we don't need the first return value from `handleNameCollision`, I think we should remove it from the functions signature.

clustercache/clustercache.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// Add the cache to a certain endpoint
+	cache.Endpoints[endpoint] = EndpointCache{
+		Expiry: time.Now().Add(cacheDuration).Format(timeLayout),
+		IDs:    c,
+	}
+
+	// Write the cache file
+	_ = write(fs, cache)
```
marians: Same as above: if the return value of `write` is not interesting, then rather make it not return anything.

marians (approved):
We spoke about a little UX improvement to be made when the name is ambiguous.

Besides that, great work! ❤️

axbarsan:
Removed the `Name` column from the name collision table. Added `Release` and `Created ` columns instead.
Updated the description with the preview

PR #511, opened 2020-03-04
Fix release lister showing certain deprecated releases as wip

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/9161

* Set the current version as wip if the previous version is active, the next version is a major upgrade and the current one is inactive


oponder (approved):
Not sure about the logic here, but it LGTM if it works for you and solves the issue.

PR #512, opened 2020-03-06
Add support for using cluster name: 'create keypair' command

axbarsan:
Towards https://github.com/giantswarm/gsctl/issues/188

This adds support for using both cluster names and IDs with the `gsctl create keypair` command.

Additional things:
* The collision confirm message has been changed to a more generic one (it was a hardcoded message regarding cluster deletion)
* Removed redudant clusterID check from the `delete cluster` command

Tested manually and is working as expected


oponder (approved):
LGTMIIWFY

PR #526, opened 2020-03-13
Fix cluster creation with wrong endpoint

axbarsan:


axbarsan:
Closed because it didn't fix the issue, nor do I have any idea what caused the issue

PR #527, opened 2020-03-25
Update apps to use v4

tfussell:
Following from https://github.com/giantswarm/gsclientgen/pull/54 and towards https://github.com/giantswarm/giantswarm/issues/8347

`gsctl` missed the versioned apps changes in https://github.com/giantswarm/api-spec/pull/192 so it needs to be updated with a newly generated `gsclientgen`.

As far as I can tell, these functions are here for https://github.com/giantswarm/giantswarm/issues/7033 but they're not really used yet so I think this change won't affect anything.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -728,9 +728,9 @@ func (w *Wrapper) GetClusterStatus(clusterID string, p *AuxiliaryParams) (*Clust
 }
 
 // CreateApp creates an App in a cluster.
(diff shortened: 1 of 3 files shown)


marians:
@tfussell Is this really related to https://github.com/giantswarm/giantswarm/issues/8347 ?

marians:
Ignore me, I got it.

PR #528, opened 2020-03-26
Enable spot instances

teemow:
* add new flags to `gsctl create nodepool`
  * `--aws-on-demand-base-capacity`: Number of on-demand instances that this node pool needs to have until spot instances are used. Default is 0
  * `--aws-spot-instance-percentage-above-base-capacity`: Percentage of spot instances used once the on-demand base capacity is fullfilled. A number of 40 would mean that 60% will be on-demand and 40% will be spot instances.
  * `--aws-use-alike-instance-types`: Use similar instance type in your node pool. This list is maintained by Giant Swarm at the moment. Eg if you select m5.xlarge then the node pool can fall back on m4.xlarge too.
* show these options in `gsctl show nodepool` and `gsctl list nodepools` too.


marians:
@teemow In the main comment you say this PR is adding a flag `--enable-spot-instances`, but in the code that's not happing. What's the plan?

Having a boolean switch to simply turn on the usage of spot instances would be completely in line with what we always did. The backend (where ever that is) should handle the rest based on sensible defaults if the user doesn't specify the rest.

marians:
Related to https://github.com/giantswarm/giantswarm/issues/8347

marians:
Could `--aws-spot-instance-percentage-above-base-capacity` be made a little shorter?

I think the `instance` part could be left out without harm, it's also not present in `--aws-on-demand-base-capacity`.

The `above-base-capacity` might also be too verbose. Assuming users will read the usage text for the explanation anyway.

It's a balance.

Here are my suggestions:

- `--aws-spot-percentage-above-base`
- `--aws-spot-percentage`

PR #530, opened 2020-04-15
Use a token defined as an environmental variable, if it exists

### LINKED ISSUE #121: There should be an environment variable to set an auth token (GSCTL_AUTH_TOKEN)
As described in https://github.com/giantswarm/gsctl/issues/47#issuecomment-332180593, endpoint selection will become available via an environment variable, as an alternative to setting it via a flag or selecting it via a command.

For consistency and security reasons, it would make sense to allow specifying an environment variable `GSCTL_AUTH_TOKEN` as an alternative to setting the `--auth-token` flag.

The security benefit would be that the token would not have to be duplicated over and over to the command line.

axbarsan:
Closes https://github.com/giantswarm/gsctl/issues/121

This gives the user the power to use an auth token set as an environmental variable.

This makes the environmental variable token behave like the token was provided as a flag. The reason for doing this in the root command, and not in the `gscliauth` config token selector, is because there is already some logic for overriding the token and using the one provided as a flag (such as defaulting the auth scheme to our GS API one).

Also, the auth token defined as an env variable has less priority than the one provided as a flag.

### Preview

image

### DIFF
--- a/commands/info/command.go
+++ b/commands/info/command.go
@@ -264,6 +264,7 @@ func getEnvironmentVariables() map[string]string {
 		"GSCTL_DISABLE_CMDLINE_TRACKING",
 		"GSCTL_DISABLE_COLORS",
 		"GSCTL_ENDPOINT",
+		"GSCTL_AUTH_TOKEN",
 		"HTTP_PROXY",
 		"KUBECONFIG",
 	}
--- a/commands/root.go
+++ b/commands/root.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/giantswarm/gscliauth/config"
 	"github.com/giantswarm/microerror"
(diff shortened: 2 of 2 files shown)


marians:
I don't find this ideal, as it overrides a flag and thus makes it look internally as if the user provided the token via the command line flag when instead it has been provided via an env variable. This is problematic in case we want to give specific feedback to end users.

Not sure if that changes things, but we might set the _default_ value of `flags.Token` to the value of the environment variable. That way we can use the [Flag](https://godoc.org/github.com/spf13/pflag#Flag) fields to determine whether the value has been set by the user or set to the default.

Also, please add the new variable to
https://github.com/giantswarm/gsctl/blob/master/commands/info/command.go#L261-L269

axbarsan:
> I don't find this ideal, as it overrides a flag and thus makes it look internally as if the user provided the token via the command line flag when instead it has been provided via an env variable. This is problematic in case we want to give specific feedback to end users.
> 
> Not sure if that changes things, but we might set the _default_ value of `flags.Token` to the value of the environment variable. That way we can use the [Flag](https://godoc.org/github.com/spf13/pflag#Flag) fields to determine whether the value has been set by the user or set to the default.
> 
> Also, please add the new variable to
> https://github.com/giantswarm/gsctl/blob/master/commands/info/command.go#L261-L269

Thank you, I've made the changes.

I think the cleaner alternative would be to add a new argument to the `ChooseToken` config function
https://github.com/giantswarm/gscliauth/blob/a3ff1611fd4849f01a957611895bc3cd87a1b954/config/config.go#L367

However, that's a rather big change that might over-complicate things. And as OIDC is right around the corner, would it be worth it?

Right now, if we really want to check if the user set the token through environmental variables or CLI flags, we can just check for the environmental variable again.

oponder:
Code looks good to me, I think the point that Marian raises is valid, but maybe we can address it when we actually get to a place where we want to give specific feedback to users.

marians (approved):

PR #531, opened 2020-04-16
Confirm TTL duration (if too long) when creating keypairs/kubeconfigs

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/4630

* Works for both `create keypair` and `create kubeconfig`
* Changed default TTL value from `30d` to `1d`
* Ask for confirmation if TTL >= 30d
* Use `--force` to bypass the TTL check

## Preview

**Validation**
image

**Force mode**
image

### DIFF
--- a/commands/create/keypair/command.go
+++ b/commands/create/keypair/command.go
@@ -9,7 +9,10 @@ import (
 	"github.com/fatih/color"
 	"github.com/giantswarm/gscliauth/config"
 	"github.com/giantswarm/gsclientgen/models"
+
 	"github.com/giantswarm/gsctl/clustercache"
+	"github.com/giantswarm/gsctl/confirm"
+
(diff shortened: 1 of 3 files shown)


oponder:
I think we should explain why the long ttl is a problem: "There is no way to revoke a keypair once it has been issued."

axbarsan:
> I think we should explain why the long ttl is a problem: "There is no way to revoke a keypair once it has been issued."

Yeah, that makes sense, I'll add it now.
Thank you 👍

oponder (approved):
LGTM :+1:

marians (approved):
Just a note: We also have API-side defaulting for the key pair TTL. But in this case I have to trouble with deriving from the backend default.

Let's make sure to communicate this default change to customers!

PR #532, opened 2020-04-16
Handle 503 error in key pair creation

marians:
Closes https://github.com/giantswarm/giantswarm/issues/10024

This makes `create kubeconfig` and `create keypair` handle the 503 error which can occur when the cluster is fairly new.

Relies on https://github.com/giantswarm/gsclientgen/pull/55

Preview

image


client/error.go
```diff
@@ -94,6 +97,9 @@ func HandleErrors(err error) {
 	if httpStatusCode == 500 {
 		headline = "An internal error occurred."
 		subtext = details
+	} else if httpStatusCode == 503 {
```
axbarsan, line 100: Not in the scope of this PR, but maybe in the future we'll start using the built-in status code constants
https://golang.org/src/net/http/status.go

axbarsan (approved):

client/error.go
marians: Done

PR #533, opened 2020-04-21
Update microerror & gscliauth modules

axbarsan:
Update the microerror & gscliauth modules, so new features could be implemented into gscliauth.


oponder (approved):
LGTM

PR #534, opened 2020-04-22
Fix SSO logout not deleting the email property from the config

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/3993

The error described in the issue was not there anymore, but the email property would not be removed from the config file.

It now is 👍


oponder (approved):
Nice, and thank you for taking on the gsctl issues! :+1:

PR #537, opened 2020-04-24
Redact auth token in telemetry report

### LINKED ISSUE #536: Auth token redaction
In the documentation I found out that you are sending every command line parameter back for statistics and other stuff. I was curious to see whether the password was redacted when you send it back. And it does: https://github.com/giantswarm/gsctl/blob/5730526c818ee0dd2fd491aa37352d7f9ca7af93/client/client.go#L186

But the redaction is only done for passwords. Not for auth tokens.

So when using this command:

```
gsctl --endpoint https://api.some.end.point.gigantic.io \
  create kubeconfig \
  --cluster xpsvd \
  --certificate-organizations system:masters \
  --auth-token dc71c38a-8600-11ea-be08-6314766e4007
```

the authentication data is still send to your telemetry server.

axbarsan:
Fixes https://github.com/giantswarm/gsctl/issues/536

Also refactored the function to accept multiple arguments/commands.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -15,7 +15,7 @@ import (
 	"github.com/giantswarm/gsctl/testutils"
 )
 
-// TestRedactPasswordArgs tests redactPasswordArgs().
+// TestRedactPasswordArgs tests redactArgs().
 func TestRedactPasswordArgs(t *testing.T) {
 	argtests := []struct {
 		in  string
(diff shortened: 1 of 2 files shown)


client/client_test.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/giantswarm/gsctl/testutils"
 )
 
-// TestRedactPasswordArgs tests redactPasswordArgs().
+// TestRedactPasswordArgs tests redactArgs().
 func TestRedactPasswordArgs(t *testing.T) {
```
marians:
```suggestion
func TestRedactArgs(t *testing.T) {
```

marians (approved):

PR #539, opened 2020-04-28
Update to go1.14

axbarsan:
Since more and more of our projects have been updated to go1.14, I figured it's time to give our old friend a newer face.

Tested locally and nothing broke 👍

PR #540, opened 2020-04-30
Replace go build version with 1.14.0

axbarsan:
Forgot this in the go1.14 upgrade PR (https://github.com/giantswarm/gsctl/pull/539)


ubergesundheit (approved):
:+1:

PR #541, opened 2020-04-30
Validate instance type before sending server request

### LINKED ISSUE #145: "create cluster" should give useful error message when bad EC2 instance type is selected
Now that we have an API endpoint to acquire information on the installation, we know which instance types are supported in an installation. See https://docs.giantswarm.io/api/#operation/getInfo

This information should be used on the client side to give feedback when a cluster is created using an instance type that is not supported.

axbarsan:
Fixes https://github.com/giantswarm/gsctl/issues/145

Even if the api provides an error if the instance type is not found, we can re-use the data we previously fetched on the client and quickly check if the instance type is available or not.

**Preview**

image


pipo02mix (approved):
that's cool LGTM

PR #543, opened 2020-05-06
Make the scale cluster command support atomic parameters

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/5578

This enables atomic modification of worker counts.

Example:
If the cluster had initially:
```
* min workers: 3
* max workers: 5
```
by only providing the `--workers-min 4` argument, only the min workers count will be changed, the max workers count will be left untouched, so the result will be:
```diff
-- * min workers: 3
++ * min workers: 4
* max workers: 5
```

### DIFF
--- a/commands/scale/cluster/command_test.go
+++ b/commands/scale/cluster/command_test.go
@@ -292,20 +292,26 @@ func TestScaleCluster(t *testing.T) {
 		initialMinScaling int
 		desiredMinScaling int
 		workersMinSet     bool
+		resultWorkersMin  int
+
 		initialMaxScaling int
 		desiredMaxScaling int
 		workersMaxSet     bool
 		numWorkersAfter   int
+		resultWorkersMax  int
 	}{
 		{
 			numWorkersBefore:  3,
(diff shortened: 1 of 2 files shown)


commands/scale/cluster/command_test.go
```diff
@@ -370,52 +425,97 @@ func TestScaleCluster(t *testing.T) {
 					]
 				}
 			}`))
-		} else {
-			w.WriteHeader(http.StatusNotFound)
-			w.Write([]byte(`{"code": "RESOURCE_NOT_FOUND", "message": "Could not find this."}`))
-		}
-	}))
-	defer mockServer.Close()
+				} else {
```
pipo02mix, line 428: is the indentation right? it is hard to judge in Github

axbarsan: Yesss it is, check it out with the blocks collapsed
image

ubergesundheit (approved):
LGTM

pipo02mix:
great, I like this addition <3

PR #544, opened 2020-05-07
Tenant cluster labeling

ubergesundheit:
Towards giantswarm/giantswarm#1693, Related to giantswarm/giantswarm#10575

This PR adds support of cluster labels described in giantswarm/docs#481 to gsctl

Changes: (Copied from giantswarm/docs#481)

- `gsctl list cluster`
  - `-o json` additionally includes user defined cluster labels
  - `--selector` or `-l` can be used to filter the list of returned clusters
- `gsctl show cluster` returns user defined cluster labels for node pool clusters on AWS ([Example](https://github.com/giantswarm/docs/pull/481/commits/d801b975b5952f437b42bb640c81d1952b12d93a#diff-5eede8a45557a2cf5bd14b48d3fdc857R42-R43))
- `gsctl create cluster`
  - cluster labels can be specified through a `labels` object in the v5 yaml cluster definition
- **new**: `gsctl update cluster`
  - `--label` allows to specify a label change in form of `key=value`. Can be specified multiple times. Set to `key=` to delete a label

### DIFF
--- a/commands/show/cluster/command.go
+++ b/commands/show/cluster/command.go
@@ -25,8 +25,6 @@ import (
 	"github.com/giantswarm/gsctl/webui"
 )
 
-const labelFilterKeySubstring = "giantswarm.io"
-
 var (
 
 	// ShowClusterCommand performs the "show cluster" function
@@ -618,7 +616,7 @@ func formatClusterLabels(labels map[string]string) []string {
 	isFirstLine := true
 
 	for key, value := range labels {
-		if strings.Contains(key, labelFilterKeySubstring) == false {
+		if strings.Contains(key, util.LabelFilterKeySubstring) == false {
 			if isFirstLine {
 				isFirstLine = false
 				formattedClusterLabels = []string{fmt.Sprintf("%s|%s=%s", color.YellowString("Labels:"), key, value)}
--- a/commands/show/cluster/command.go
+++ b/commands/show/cluster/command.go
@@ -25,6 +25,8 @@ import (
 	"github.com/giantswarm/gsctl/webui"
 )
 
+const labelFilterKeySubstring = "giantswarm.io"
+
 var (
 
 	// ShowClusterCommand performs the "show cluster" function
@@ -616,7 +618,7 @@ func formatClusterLabels(labels map[string]string) []string {
 	isFirstLine := true
 
 	for key, value := range labels {
-		if strings.Contains(key, "giantswarm.io") == false {
+		if strings.Contains(key, labelFilterKeySubstring) == false {
 			if isFirstLine {
 				isFirstLine = false
 				formattedClusterLabels = []string{fmt.Sprintf("%s|%s=%s", color.YellowString("Labels:"), key, value)}
--- a/commands/show/cluster/command.go
+++ b/commands/show/cluster/command.go
@@ -492,6 +492,7 @@ func printV5Result(args Arguments, details *models.V5ClusterDetailsResponse,
 	clusterTable = append(clusterTable, color.YellowString("Kubernetes API endpoint:")+"|"+details.APIEndpoint)
 	clusterTable = append(clusterTable, color.YellowString("Master availability zone:")+"|"+details.Master.AvailabilityZone)
 	clusterTable = append(clusterTable, color.YellowString("Release version:")+"|"+details.ReleaseVersion)
+	clusterTable = append(clusterTable, formatClusterLabels(details.Labels)...)
 
 	// BYOC credentials.
 	if credentialDetails != nil && credentialDetails.ID != "" {
@@ -607,4 +608,23 @@ func formatNodePoolDetails(nodePools *models.V5GetNodePoolsResponse) []string {
 	rows = append(rows, color.YellowString("RAM in nodes (GB):")+fmt.Sprintf("|%d", ramGB))
 
 	return rows
+}
+
+func formatClusterLabels(labels map[string]string) []string {
+	formattedClusterLabels := []string{color.YellowString("Labels:|") + "-"}
+
+	isFirstLine := true
+
+	for key, value := range labels {
+		if strings.Contains(key, "giantswarm.io") == false {
+			if isFirstLine {
+				isFirstLine = false
+				formattedClusterLabels = []string{fmt.Sprintf("%s|%s=%s", color.YellowString("Labels:"), key, value)}
+			} else {
+				formattedClusterLabels = append(formattedClusterLabels, fmt.Sprintf("|%s=%s", key, value))
+			}
+		}
+	}
+
+	return formattedClusterLabels
 }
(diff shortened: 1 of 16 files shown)


ubergesundheit:
Oops. pinged too early

commands/show/cluster/command.go
```diff
@@ -608,3 +609,22 @@ func formatNodePoolDetails(nodePools *models.V5GetNodePoolsResponse) []string {
...
 	return rows
 }
+
+func formatClusterLabels(labels map[string]string) []string {
+	formattedClusterLabels := []string{color.YellowString("Labels:|") + "-"}
+
+	isFirstLine := true
+
+	for key, value := range labels {
+		if strings.Contains(key, "giantswarm.io") == false {
```
axbarsan, line 619: Can you please extract the `giantswarm.io` token into a constant at the top of the file?

commands/update/cluster/command.go
```diff
@@ -56,13 +57,15 @@ func init() {
 func initFlags() {
 	Command.ResetFlags()
 	Command.Flags().StringVarP(&flags.Name, "name", "n", "", "new cluster name")
+	Command.Flags().StringSliceVar(&flags.Label, "label", nil, "modification of a label in form of key=value. Can be specified multiple times. To delete a label set to key=")
```
axbarsan, line 60:
```suggestion
	Command.Flags().StringSliceVar(&flags.Label, "label", nil, "modification of a label in form of 'key=value'. Can be specified multiple times. To delete a label set to 'key='")
```

commands/update/cluster/command.go
```diff
@@ -226,4 +284,15 @@ func printResult(cmd *cobra.Command, positionalArgs []string) {
 	}
 
 	fmt.Println(color.GreenString("Cluster '%s' has been modified.", arguments.ClusterNameOrID))
+	if result.ClusterName != "" {
+		fmt.Printf("New cluster name: '%s'\n", result.ClusterName)
+	}
+	if len(result.Labels) > 0 {
+		fmt.Println("New cluster labels:")
+		for key, label := range result.Labels {
+			if strings.Contains(key, "giantswarm.io") == false {
```
axbarsan, line 293: Same here with the `giantswarm.io` constant

axbarsan (commented):
Looks good to me overall, just a few nitpicks

commands/update/cluster/command.go
```diff
@@ -18,20 +18,23 @@ import (
 	"github.com/giantswarm/gsctl/flags"
 )
 
+const labelFilterKeySubstring = "giantswarm.io"
```
MarcelMue, line 21: Not a big deal but is there a place where this can be defined once instead of in every command?

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/giantswarm/apiextensions v0.0.0-20191213075442-71155aa0f5b7
 	github.com/giantswarm/columnize v2.0.3-0.20190718092621-cc99d98ffb29+incompatible
 	github.com/giantswarm/gscliauth v0.2.1
-	github.com/giantswarm/gsclientgen v0.0.0-20200423130825-3202ff18d14f
+	github.com/giantswarm/gsclientgen v1.0.2-0.20200507100632-8ff496037482
```
MarcelMue, line 13: Reminder to use a properly tagged version here.

commands/update/cluster/command_test.go
```diff
@@ -302,3 +336,43 @@ func TestExecuteWithError(t *testing.T) {
 		})
 	}
 }
+
+func Test_modifyClusterLabelsRequestFromArguments(t *testing.T) {
```
MarcelMue, line 340: I would like to use tabletests more often but understand that it can be cumbersome here. So unit tests is still be than no unit tests.

MarcelMue (approved):
Looks fine to me - I assume you tested this fun stuff.

axbarsan (approved):

commands/update/cluster/command.go
```diff
@@ -130,9 +143,21 @@ func printValidation(cmd *cobra.Command, positionalArgs []string) {
...
+	if args.Name == "" {
+		return updateLabels(args)
+	}
+
+	return nil, microerror.Mask(errors.NoOpError)
+}
+
+// updateName updates the cluster.
+// It determines whether it is a v5 or v4 cluster and uses the appropriate mechanism.
+func updateName(args Arguments) (*result, error) {
```
stone-z, line 160: Maybe consider aligning the comments on `updateName` and `updateLabels`. Is the label behavior v5 only?

stone-z (approved):
Minor comment. Otherwise LGTM

go.mod
ubergesundheit: This is the result of `go get github.com/giantswarm/gsclientgen@8ff4960` (which is also tagged `v2.0.4`)

Running the command above with `@v2.0.4` fails with: `go get github.com/giantswarm/gsclientgen@v2.0.4: github.com/giantswarm/gsclientgen@v2.0.4: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2`

I think I need a little pointer on how to resolve this

axbarsan: I believe you need to use `github.com/giantswarm/gsclientgen/v2` in the code to be able to have the v2 version in go.mod

commands/update/cluster/command.go
ubergesundheit: yeah labels are v5 only. I guess I'll add a check like its present in `updateName` to abort the action for v4 clusters :+1:

go.mod
ubergesundheit: Tried that. Does not work. I've replaced all `github.com/giantswarm/gsclientgen` with `github.com/giantswarm/gsclientgen/v2` in all `*.go` files, then ran `go run main.go`.

Errors with `... module github.com/giantswarm/gsclientgen@latest found (v2.0.3+incompatible), but does not contain package github.com/giantswarm/gsclientgen/v2/client`

Does the [go.mod](https://github.com/giantswarm/gsclientgen/blob/master/go.mod) file of gsclientgen need to specify the same import path? `github.com/giantswarm/gsclientgen/v2` ?

axbarsan: https://github.com/giantswarm/gsclientgen/blob/8ff49603748206bdea4ab8f548d7702557a141b2/go.mod#L1

This should be `module github.com/giantswarm/gsclientgen/v2` AFAIK

ubergesundheit: Tried that (https://github.com/giantswarm/gsclientgen/commit/780706a7448880f905e85c0f2610f2df8176b6f1), but with the changed import paths in all files, parameters etc are not compatible any more. I'm guessing all import paths in go files of gsclientgen need to be adapted? This seems like an antipattern since gsclientgen is generated.

Or is there something I am missing?

axbarsan: Yeah, you're right. We need to find a way to add versioning to the client generation, or else we will not be able to use released versions

ubergesundheit: So for now If its okay for you @MarcelMue @axbarsan lets keep this like it is and address the changes in gsclientgen separate?

MarcelMue: Fine for now but yeah let's take care.

PR #547, opened 2020-05-08
Revert "Validate instance type before sending server request"

axbarsan:
Reverts giantswarm/gsctl#541

With https://github.com/giantswarm/gsctl/pull/546 , we no longer need any client-side validation. The API will handle and we will display the API error, which displays what the problem is.


oponder (approved):
LGTM! nice

PR #548, opened 2020-05-08
Show detailed output when ping fails

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/10634

## Preview

image


ubergesundheit (approved):
LGTM

PR #549, opened 2020-05-08
Fix app not working in alpine linux

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/10506

Tested in both containers, and in an Ubuntu VM. Everything's working as expected. 👍


ubergesundheit (approved):
Awesome!

marians:
Do you want to add a comment that CGO is disabled on purpose to enable running in Docker containers, including those based on Alpine?

PR #550, opened 2020-05-13
Deprecate 'gsctl version' command and use '--version' flag instead

### LINKED ISSUE #172: Use cobra's built in --version command as a replacement for 'gsctl version'
Cobra supports a --version flag in its latest version. Since many commands use cobra and `program --version` might be almost a convention, we should consider moving the function of `gsctl version` to `gsctl --version`.

axbarsan:
Fixes https://github.com/giantswarm/gsctl/issues/172

This PR deprecates `gsctl version` (it won't show up in the help command, and it will show a warning message if you use it). It will still work , until we're ready to remove it completely.

## Preview (with placeholder versions)

**`$ gsctl version`**

image

**`$ gsctl --version`**

image

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -44,6 +44,7 @@ var RootCommand = &cobra.Command{
 	Use: config.ProgramName,
 	// this is inherited by all child commands
 	PersistentPreRunE: initConfig,
+	Run:               printResult,
 }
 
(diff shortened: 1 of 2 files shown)


pipo02mix (approved):
nice

PR #551, opened 2020-05-14
Custom sort parameters for the 'list clusters' command

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/10811

## Changes

**generic**
* Created a generic `sortable` package that can be used for sorting stuff
* Created a custom `table` package that can be used for displaying data in a cool table, but that also supports sorting the data

**list clusters**
* Refactored the command to use the new table
* The command now accepts the `-s/--sort` flag, that works with any table column initials or part of words (doesn't matter in which casing)
  * If there are multiple columns matching with that token, an error is thrown
  * If there is no column matching with that token, an error is thrown
* Same sorting logic is adapted and applied to json output

## Help 

**Examples**

image

**Flags**

image

## Possible errors

**Attribute not found**

image

**Multiple attributes found for token**

image

### DIFF
--- a/commands/list/clusters/command.go
+++ b/commands/list/clusters/command.go
@@ -316,35 +316,35 @@ func createTable(args Arguments) *table.Table {
 			Name:        tableColID,
 			DisplayName: "ID",
 			Sortable: sortable.Sortable{
-				SortType: sortable.Types.String,
+				SortType: sortable.String,
 			},
 		},
 		{
 			Name:        tableColOrg,
 			DisplayName: "ORGANIZATION",
 			Sortable: sortable.Sortable{
-				SortType: sortable.Types.String,
+				SortType: sortable.String,
 			},
 		},
 		{
 			Name:        tableColName,
 			DisplayName: "NAME",
 			Sortable: sortable.Sortable{
-				SortType: sortable.Types.String,
+				SortType: sortable.String,
 			},
 		},
 		{
 			Name:        tableColRelease,
 			DisplayName: "RELEASE",
 			Sortable: sortable.Sortable{
-				SortType: sortable.Types.Semver,
+				SortType: sortable.Semver,
 			},
 		},
 		{
 			Name:        tableColCreateDate,
 			DisplayName: "CREATED",
 			Sortable: sortable.Sortable{
-				SortType: sortable.Types.Date,
+				SortType: sortable.Date,
 			},
 		},
 	}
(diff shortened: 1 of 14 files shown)


marians (changes requested):
Impressive work! 🚀 

Some wishes for improvements:

- The sorting by cluster name and org should use case-insensitive [alphanumerical](http://davekoelle.com/alphanum.html) sorting. There is an implementation of the alphanumerical algorithm in https://github.com/facette/natsort
- The flag info `-s, --sort string       Sort by the given field` should contain the field names that can be used to sort by. Suggestion: `Sort by one of the fields id, o(rganization), n(ame), r(elease), c(reated), d(eleting-since)
- I noticed we don't have any examples in the usage for this command. Could you please add an example using both `--sort` and `--output json`? Potentially also `--selector`.

---

When using `--sort del` without using `--show-deleting`, this error occurs:

```
./gsctl list clusters -e ginger --sort del
Cannot sort by attribute 'del'.
The attribute 'del' does not exist.
You can sort by any of these attributes: id, created, name, organization, release, deleting-since
```

The fact that `deleting-since` appears in the list of possible sort attributes is misleading.

axbarsan:
> * The sorting by cluster name and org should use case-insensitive [alphanumerical](http://davekoelle.com/alphanum.html) sorting. There is an implementation of the alphanumerical algorithm in https://github.com/facette/natsort

We're now using that algorithm for sorting strings. Thank you for the great find 👍 

> * The flag info `-s, --sort string       Sort by the given field` should contain the field names that can be used to sort by. Suggestion: `Sort by one of the fields id, o(rganization), n(ame), r(elease), c(reated), d(eleting-since)

✅ (I printed `id` as `i(d)` because using `--sort i` works.

> * I noticed we don't have any examples in the usage for this command. Could you please add an example using both `--sort` and `--output json`? Potentially also `--selector`.

✅ 

> When using `--sort del` without using `--show-deleting`, this error occurs:
> 
> ```
> ./gsctl list clusters -e ginger --sort del
> Cannot sort by attribute 'del'.
> The attribute 'del' does not exist.
> You can sort by any of these attributes: id, created, name, organization, release, deleting-since
> ```
> 
> The fact that `deleting-since` appears in the list of possible sort attributes is misleading.

✅ I added the possibility to hide columns, and sorting will not throw any errors even if that column is hidden and the data is not there

commands/list/clusters/command.go
```diff
@@ -32,9 +32,22 @@ var (
...
+
+  gsctl list clusters
+
+  gsctl list clusters --output json
+
+  gsctl list clusters --show-deleting
+
+  gsctl list clusters --selector environment=testing
+
+  gsctl list clusters --sort organization
```
marians, line 47:
```suggestion
  gsctl list clusters --sort org
```

Let's hunch users to save some typing :)

marians (approved):
❤️

PR #554, opened 2020-05-25
Implement AWS HA Masters

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/10787

### Create cluster

* **Release does not support HA Masters**

image

* **Using both the 'master' key and the 'master_nodes' key**

image

* **Using the 'master' key on a release that supports HA masters**

image

### Show cluster

* **No AZ information, 1 Master, 0 Masters ready**

image

* **3 AZs, 3 Masters, 3 Masters ready**

image

* **Older release, doesn't support HA masters**

image

### Update cluster

* **Trying to switch to a single master node**

image

* **Trying to switch to HA a cluster that has an older release**

image


commands/create/cluster/command.go
```diff
@@ -51,23 +52,30 @@ type Arguments struct {
...
 }
 
 // collectArguments gets arguments from flags and returns an Arguments object.
-func collectArguments() Arguments {
+func collectArguments(cmd *cobra.Command) Arguments {
 	endpoint := config.Config.ChooseEndpoint(flags.APIEndpoint)
 	token := config.Config.ChooseToken(endpoint, flags.Token)
 	scheme := config.Config.ChooseScheme(endpoint, flags.Token)
 
+	var haMasters *bool
```
axbarsan, line 66: If the flag has been set by the user, then we'll actually set the value... Otherwise, leave it nil

commands/create/cluster/v5.go
```diff
@@ -4,30 +4,59 @@ import (
...
-		def.Owner = owner
+	if flags.owner != "" {
+		def.Owner = flags.owner
+	}
+
+	if haMastersFeature.HAMasters.IsSupported(config.Config.Provider, def.ReleaseVersion) || def.ReleaseVersion == "" {
+		var isHA bool
+
+		if flags.isHAMaster != nil {
+			isHA = *flags.isHAMaster
```
axbarsan, line 47: If the flag is set by the user

commands/create/cluster/v5.go
```diff
@@ -4,30 +4,59 @@ import (
...
+		} else {
+			if def.MasterNodes != nil {
```
axbarsan, line 49: If master nodes are set in the YAML cluster definition

commands/create/cluster/v5.go
```diff
@@ -4,30 +4,59 @@ import (
...
+				isHA = def.MasterNodes.HighAvailability
+			} else {
+				// Default to true if there was no value set.
+				isHA = true
```
axbarsan, line 53: Default to true otherwise

commands/create/cluster/v5.go
```diff
@@ -88,8 +123,16 @@ func addClusterV5(def *types.ClusterDefinitionV5, args Arguments, clientWrapper
 		return "", true, microerror.Mask(errors.ClusterOwnerMissingError)
 	}
 
+	if def.MasterNodes != nil && def.ReleaseVersion != "" && !haMastersFeature.HAMasters.IsSupported(config.Config.Provider, def.ReleaseVersion) {
```
axbarsan, line 126: If the user set a release version that is not supported, and also set HA Master, throw an error

commands/update/cluster/command.go
```diff
@@ -90,20 +98,20 @@ func collectArguments(positionalArgs []string) Arguments {
...
-	}
-	if args.AuthToken == "" && args.UserProvidedToken == "" {
+	} else if args.AuthToken == "" && args.UserProvidedToken == "" {
 		return microerror.Mask(errors.NotLoggedInError)
 	} else if args.ClusterNameOrID == "" {
 		return microerror.Mask(errors.ClusterNameOrIDMissingError)
-	} else if args.Name == "" && (args.Labels == nil || len(args.Labels) == 0) {
-		return microerror.Mask(errors.NoOpError)
-	} else if args.Name != "" && args.Labels != nil && len(args.Labels) != 0 {
+	} else if cmd.Flag("master-ha").Changed && !args.MasterHA {
```
axbarsan, line 112: If the user tried to disable HA master

pkg/featuresupport/featuresupport.go
```diff
@@ -0,0 +1,53 @@
+package featuresupport
```
axbarsan, line 1: I created this package to easily provide different features for different providers and various release versions.

pkg/featuresupport/hamasters/aws.go
```diff
@@ -0,0 +1,12 @@
+package hamasters
+
+import (
+	"github.com/giantswarm/gsctl/pkg/featuresupport"
+)
+
+var (
+	aws = featuresupport.Provider{
+		Name:            "aws",
+		RequiredVersion: "11.5.0",
```
marians, line 10: This information should be taken from the API, just like with node pools. The api-spec branch adds a feature named `ha_masters`. See also https://github.com/giantswarm/api/blob/hamasters/server/endpoint/v4/info/response/features.go#L7

commands/create/cluster/command.go
```diff
@@ -193,14 +203,15 @@ func initFlags() {
 	Command.Flags().StringVarP(&flags.ClusterName, "name", "n", "", "Cluster name")
 	Command.Flags().StringVarP(&flags.Owner, "owner", "o", "", "Organization to own the cluster")
 	Command.Flags().StringVarP(&flags.Release, "release", "r", "", "Release version to use, e. g. '1.2.3'. Defaults to the latest. See 'gsctl list releases --help' for details.")
+	Command.Flags().BoolVar(&cmdMasterHA, "master-ha", true, "This means the cluster will have three master nodes. Requires High-Availability Master support.")
```
marians, line 206: I think `cmdMasterHA` should be defined in the `flags` package, just like the others.

commands/create/cluster/command.go
```diff
@@ -239,6 +250,18 @@ func printResult(cmd *cobra.Command, positionalArgs []string) {
 		richError, richErrorOK := err.(*errgo.Err)
 
 		switch {
+		case IsHAMastersNotSupported(err):
+			headline = "Feature not supported"
+
+			requiredVersion := haMastersFeature.HAMasters.RequiredVersion(config.Config.Provider)
+			if requiredVersion != nil {
+				subtext = fmt.Sprintf("HA Masters is only supported by releases newer than %s.", *requiredVersion)
```
marians, line 258:
```suggestion
				subtext = fmt.Sprintf("Master node high availability is only supported by releases %s and higher.", *requiredVersion)
```

commands/create/cluster/command.go
```diff
@@ -239,6 +250,18 @@ func printResult(cmd *cobra.Command, positionalArgs []string) {
...
+			} else {
+				subtext = "HA Masters is not supported by your provider."
```
marians, line 260:
```suggestion
				subtext = "Master node high availability is not supported on this provider."
```

Can we also add the provider name to the error message? I am not sure every user would know what the provider is in this context. Like `Master node high availability is not supported on this provider (KVM).`

commands/create/cluster/command.go
```diff
@@ -239,6 +250,18 @@ func printResult(cmd *cobra.Command, positionalArgs []string) {
...
+			}
+		case IsMustProvideSingleMasterType(err):
+			headline = "Conflicting master node configuration"
+			subtext = "The release version you're trying to use supports HA Masters.\nPlease remove the 'master' field from your cluster definition and use the '--master-ha' flag."
```
marians, line 264:
```suggestion
			subtext = "The release version you're trying to use supports master node high availability.\nPlease remove the 'master' attribute from your cluster definition and use the 'master_nodes' attribute instead."
```

pkg/featuresupport/hamasters/aws.go
axbarsan: You killed all my hopes and dreams :(

capabilities/service.go
```diff
@@ -59,13 +60,23 @@ func (s *Service) initCapabilities() error {
...
+	// Enhance feature info for HA Masters, if available.
+	if info.Payload.Features != nil && info.Payload.Features.HaMasters != nil {
+		HAMasters.RequiredReleasePerProvider = []ReleaseProviderPair{
+			{
+				Provider:       info.Payload.General.Provider,
+				ReleaseVersion: semver.MustParse(info.Payload.Features.HaMasters.ReleaseVersionMinimum),
+			},
+		}
+	}
+
```
marians, line 79: I think this would become more readable with a common block like

```
if info.Payload.Features != nil {
    if info.Payload.Features.Nodepools != nil { ... }
    if info.Payload.Features.HaMasters != nil { ... }
}
```

axbarsan:
Updated screenshots and added test for validating HA master nodes vs single masters

axbarsan:
Added support for only printing the `master nodes ready` counter if the cluster version supports it.

Screenshot at `Show cluster`:`Older release, doesn't support HA masters`

marians:
BTW, wanna mark this ready for review?

marians:
Will do some more end 2 end testing today

marians (approved):

PR #563, opened 2020-06-16
Account for empty AZ labels in node pools lister

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/11150

We were slicing a string with the starting index of `-1` if the string was empty.


marians (approved):
LGTM if it works for you

PR #564, opened 2020-06-17
Extend cluster definition to support spot instances

marians:
The goal of this PR is to add to `gsctl create cluster` the missing support for node pool attributes that got added witt spot instances, in the case that a YAML cluster definition is used.

PR #565, opened 2020-06-19
Fix node pool creation based on cluster definition

marians:
Follow-up to https://github.com/giantswarm/gsctl/pull/564

This ensures that spot-related attributes are applied when creating node pools based on a cluster definition.

It also fixes a misunderstanding regarding `on_demand_percentage_above_base_capacity` in unit tests, which I introduced in the last PR.


axbarsan (approved):
LGTM

PR #567, opened 2020-07-13
Azure v5 list nodepools

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/9

What's in the box:
- Support for listing Azure node pools


commands/list/nodepools/command.go
```diff
@@ -222,6 +217,33 @@ func getOutput(nps []*models.V5GetNodePoolsResponseItems, outputFormat string) (
...
+		if err != nil {
+			return "", microerror.Mask(err)
+		}
+	} else if np.NodeSpec.Azure != nil && np.NodeSpec.Aws == nil {
+		output, err = getOutputAzure(nps)
+		if err != nil {
+			return "", microerror.Mask(err)
+		}
+	} else {
+		return "", microerror.Mask(errors.ClusterDoesNotSupportNodePoolsError)
```
nprokopic, line 235: I'm just not sure what to display if we cannot figure out which provider is this (which would mean that there is an issue in the API). An error? Or maybe only show fields that are common across all providers?

tuommaki: `fetchNodepools()` already does this check (and returns same error). I think this is enough here.

marians:
Once the command is implemented, a few screenshots for demonstration would be appreciated. Also, is there a docs PR we can look at in parallel?

commands/list/nodepools/command.go
```diff
@@ -296,6 +318,74 @@ func getOutput(nps []*models.V5GetNodePoolsResponseItems, outputFormat string) (
...
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	headers := []string{
+		color.CyanString("ID"),
+		color.CyanString("NAME"),
+		color.CyanString("AZ"),
+		color.CyanString("VM SIZE"),
+		// color.CyanString("NODES MIN/MAX"),
```
marians, line 332: Will this be used or not? We shouldn't merge it commented out.

nprokopic: We (Celestial) had some discussions about the topic and autoscaling is going be implemented probably after the node pools, so I'll remove this for now.

nprokopic:
> Once the command is implemented, a few screenshots for demonstration would be appreciated.

It is not implemented yet on the operators side, `api`/`gsctl` are tested with manually created CRs, but there are no actual node pools / machines deployed yet. For example it always shows zero nodes and no CPU and RAM:
```
$ ./build/bin/gsctl-linux-amd64 list nodepools np001
ID       NAME     AZ  VM SIZE          NODES DESIRED  NODES READY  CPUS  RAM (GB)
np001-1  np001-1  2   Standard_D2s_v3              0            0     0       0.0

$ ./build/bin/gsctl-linux-amd64 list nodepools np002
ID       NAME     AZ   VM SIZE          NODES DESIRED  NODES READY  CPUS  RAM (GB)
np002-1  np002-1  1    Standard_D2s_v3              0            0     0       0.0
np002-2  np002-2  2,3  Standard_D2s_v3              0            0     0       0.0
```

> Also, is there a docs PR we can look at in parallel?

There is not at the moment. Should we update docs even though Azure node pools are not actually implemented yet?
I guess we would update these pages:
- https://github.com/giantswarm/docs/blob/master/src/content/reference/gsctl/list-nodepools.md
- https://github.com/giantswarm/docs/blob/master/src/content/reference/cluster-definition/_index.md#node-pool-definition-keys-v5-nodepool-keys

nodespec/azure.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// GetVMSizeDetails returns info on a certain VM size
+func (p *ProviderAzure) GetVMSizeDetails(name string) (*VMSize, error) {
+	vmSize, ok := p.vmSizes[name]
+	if ok {
+		return &vmSize, nil
+	}
+
+	fmt.Printf("VM size not found %q", name)
```
axbarsan, line 260: We usually keep error printing into a single place, as it's much easier to deal with if you want to change something. You can handle the error printing here, instead
https://github.com/giantswarm/gsctl/blob/30bf71c831ca4640523cf7a7180a257ebf22ccd4/commands/list/nodepools/command.go#L386

nprokopic: Damn, this was a debugging leftover, thanks for pointing out :)

nodespec/azure.go
```diff
@@ -0,0 +1,263 @@
...
+	NumberOfCores        int64   `yaml:"numberOfCores"`
+	OSDiskSizeInMB       int64   `yaml:"osDiskSizeInMb"`
+	ResourceDiskSizeInMB float64 `yaml:"resourceDiskSizeInMb"`
+}
+
+// NewAzureProvider initiates a new Azure provider with the information about VM sizes.
+func NewAzureProvider() (*ProviderAzure, error) {
+	p := &ProviderAzure{}
+
+	err := yaml.Unmarshal([]byte(azureVMSizesYAML), &p.vmSizes)
```
axbarsan, line 245: I wonder if we actually need to store them in YAML. I mean, it's easy to just copy/paste the VM Size from the installations repo, but it's very fragile...
Wouldn't we be better off storing them in a `map[string]VMSize`? It would definitely speed things up, since we wouldn't have to unmarshal the YAML

axbarsan:
As for docs, I'd rather keep a bigger open PR for all the Azure V5 changes, where we can add all the required changes, and which we can merge when the feature is ready

nodespec/azure.go
nprokopic: I just went the AWS-way as it was the quickest :grin:

Manually translating YAML to `map[string]VMSize` could be error prone maybe. Ideally we would fetch VM sizes data from a single place, wherever we need it.

nprokopic:
@marians does it work for you to have single docs pull request for all Azure V5 / node pool changes as Adrian suggested? It would indeed be simpler/easier than updating docs separately for every part of node pools implementation.

marians:
Sure, works for me

PR #568, opened 2020-07-20
Azure v5: create node pool

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12223

⚠️  Asking for review, but not merging until API is in master.

It's supposed to behave exactly like the AWS implementation, with a few additions:

* Updated command usage information
https://github.com/giantswarm/gsctl/blob/14851c38d8e07d2df1a6f382f8d83c3e59c3702e/commands/create/nodepool/command.go#L33

* Using both `--aws-instance-type` and `--azure-vm-size` flags throws an error
image

* Using invalid format for availability zones throws an error
image


pipo02mix (approved):
LGTM

PR #575, opened 2020-08-06
json output to create cluster, create kubeconfig, delete cluster

ubergesundheit:
Closed in favour of #580 

This PR adds handling of `--output=json` flag `create cluster`, `create kubeconfig` and `delete cluster` used by [`standup`](https://github.com/giantswarm/standup).

**`gsctl create cluster`**

Success output is `{"id":"asd91","result":"created"}`. Field `result` can be `created`, `created-with-errors`, `error`. 

In case of `"result": "error"` an extra `error` field will contain the marshalled error.

**`gsctl create kubeconfig`**

Success output is `{"result":"ok","kubeconfig":"apiVersion: v1\nkind: Config\nclusters:\n- name: ......."}`. Field `result` can be `ok`, `error`.

In case of `"result": "error"` an extra `error` field will contain the marshalled error.

**`gsctl delete cluster`**

Success output is  `{"result":"deletion scheduled","id":"jkp0v"}`. Field `result` can be `deletion scheduled`, `error`.

In case of `"result": "error"` an extra `error` field will contain the marshalled error.

**Summary of other changes:**

- Use updated `gscliauth` library to handle `--endpoint` and `--auth-token` flags without needing to `gsctl login` first (Changes from [this PR](https://github.com/giantswarm/gscliauth/pull/22))


marians:
How would gsctl behave in case a command is called with `--output=json` and JSON oputput is not supported in the command? I think it should at leat print a warning/error and exit with a code > 0.

ubergesundheit:
Good point, right now it does not change anything for commands not changed for the global flag. Your suggestion will be included in the final version of this PR :+1: 

But since this is just a hacking prototype, there it will probably change significantly.

commands/create/kubeconfig/command.go
```diff
@@ -344,6 +360,17 @@ func createKubeconfigRunOutput(cmd *cobra.Command, cmdLineArgs []string) {
 
 	// Success output
 
+	if arguments.outputFormat == formatting.OutputFormatJSON {
+		jsonResult := JSONOutput{Result: "ok", KubeConfig: string(result.selfContainedYAMLBytes)}
+		outputBytes, err := json.MarshalIndent(jsonResult, formatting.OutputJSONPrefix, formatting.OutputJSONIndent)
+		if err != nil {
+			os.Exit(1)
```
stone-z, line 367: Could use a message here as well

commands/delete/cluster/command.go
```diff
@@ -142,6 +154,9 @@ func validatePreconditions(args Arguments) error {
 	if config.Config.Token == "" && args.token == "" {
 		return microerror.Mask(errors.NotLoggedInError)
 	}
+	if args.outputFormat != "" && args.outputFormat != formatting.OutputFormatJSON {
+		return microerror.Maskf(errors.OutputFormatInvalidError, fmt.Sprintf("Output format '%s' is unknown. Valid options: '%s'", args.outputFormat, formatting.OutputFormatJSON))
```
stone-z, line 158: Same here

commands/delete/cluster/command.go
```diff
@@ -176,6 +191,18 @@ func printResult(cmd *cobra.Command, args []string) {
 		if arguments.clusterNameOrID != "" {
 			clusterID = arguments.clusterNameOrID
 		}
+
+		if arguments.outputFormat == formatting.OutputFormatJSON {
+			jsonResult := JSONOutput{Result: "deletion scheduled", ID: clusterID}
+			outputBytes, err := json.MarshalIndent(jsonResult, formatting.OutputJSONPrefix, formatting.OutputJSONIndent)
+			if err != nil {
+				os.Exit(1)
```
stone-z, line 199: Message here maybe

stone-z (commented):
Some notes

commands/create/cluster/command.go
```diff
@@ -102,6 +106,16 @@ type creationResult struct {
 	HasErrors bool
 }
 
+// JSONOutput contains the fields included in JSON output of the create cluster command when called with json output flag
+type JSONOutput struct {
+	// cluster ID
```
tfussell, line 111: We don't really enforce Go Doc comments where the identifier is the first word of the comment, but would be nice to stick to it for consistency here: like maybe `ID of the cluster`. Same with `Result` below.

axbarsan:
Please don't forget to add your changes to the `gsctl` docs when you find the time 👍

ubergesundheit:
> Please don't forget to add your changes to the `gsctl` docs when you find the time +1

Yeah I'll do before creating an release!

Closing this one in favour of #580 as the other one has coveralls green for the same changes ¯\_(ツ)_/¯

PR #579, opened 2020-08-20
Assorted changes spliced from add-json-output PR

ubergesundheit:
These are the unrelated changes from #575 

- Create formatting/output.go file containing output flag related constants
- Make all commands having an output flag use the constants in formatting/output.go
- Some fixes to tests


axbarsan (approved):
Me likey

PR #581, opened 2020-08-26
azure v5: NP lister: show placeholder for unknown VM sizes/instance types

axbarsan:
As seen on https://gigantic.slack.com/archives/CNUDS3AMU/p1598437605023500

We already had something like this for the `show nodepool` command, but not here.

### Before 

image

### After

image


tuommaki (approved):
LGTM

PR #587, opened 2020-09-14
Use buster golang image for releases in Makefile

ubergesundheit:
Apparently the `-stretch` image of go 1.15 does not exist on the docker hub. This PR switches to the `-buster` image for creating release binaries.


marians (approved):
LGTM if it works for you

PR #590, opened 2020-09-17
Validate if node pool is actually changed before trying to update it

axbarsan:
This validates that the node pool is actually changed before trying to send an update request. If it isn't updated, it throws this error:

image

### DIFF
--- a/commands/update/nodepool/command.go
+++ b/commands/update/nodepool/command.go
@@ -165,6 +165,86 @@ func printValidation(cmd *cobra.Command, positionalArgs []string) {
 		return
 	}
 
+	handleError(err)
+	os.Exit(1)
+}
+
+func updateNodePool(args Arguments) (*result, error) {
+	clientWrapper, err := client.NewWithConfig(args.APIEndpoint, args.UserProvidedToken)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


commands/update/nodepool/command.go
```diff
@@ -194,29 +175,40 @@ func updateNodePool(args Arguments) (*result, error) {
...
 
-	response, err := clientWrapper.ModifyNodePool(clusterID, args.NodePoolID, requestBody, auxParams)
+	updated := false
+	modifyRequestBody := &models.V5ModifyNodePoolRequest{}
+	{
+		if args.Name != "" && args.Name != existingNP.Payload.Name {
+			modifyRequestBody.Name = args.Name
+			updated = true
+		}
+		if args.ScalingMin != 0 && args.ScalingMin != existingNP.Payload.Scaling.Min ||
```
fiunchinho, line 197: is the name and the scaling values the only things that can be updated? What about vmsize?

axbarsan: Currently, that can't be updated
https://docs.giantswarm.io/api/#operation/modifyNodePool

fiunchinho (approved):

PR #596, opened 2020-10-14
Don t encourage the usage of pre release versions

axbarsan:
With these changes:
* When creating a new cluster without specifying a release version, the default version used will **NOT** be a pre-release version (or a version with custom build metadata)
* When upgrading a cluster, the version picked automatically by the application will **NOT** be a pre-release version (or a version with custom build metadata)

### DIFF
--- a/commands/upgrade/cluster/command.go
+++ b/commands/upgrade/cluster/command.go
@@ -5,12 +5,14 @@ import (
 	"os"
 	"sort"
 
+	"github.com/Masterminds/semver"
 	"github.com/fatih/color"
 	"github.com/giantswarm/gscliauth/config"
 	"github.com/giantswarm/gsclientgen/v2/models"
-	"github.com/giantswarm/gsctl/clustercache"
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/gsctl/clustercache"
 
 	"github.com/giantswarm/gsctl/client"
 	"github.com/giantswarm/gsctl/client/clienterror"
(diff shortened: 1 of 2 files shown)


commands/upgrade/cluster/command.go
```diff
@@ -359,6 +361,15 @@ func upgradeCluster(args Arguments) (*upgradeClusterResult, error) {
 	return result, nil
 }
 
+func isVersionProductionReady(version string) bool {
+	semverVersion, err := semver.NewVersion(version)
+	if err != nil {
+		return false
+	}
+
+	return len(semverVersion.Prerelease()) < 1 && len(semverVersion.Metadata()) < 1
```
pipo02mix, line 370: What this `semverVersion.Metadata()` tell you?

marians: https://pkg.go.dev/github.com/Masterminds/semver/v3#Version.Metadata

It probably returns `-alpha1` for a string like `13.0.0-alpha1`.

marians: Or rather `alpha1` instead of `-alpha1`

marians: Well, here is an example: https://play.golang.org/p/Bbvs1yIxAn8

pipo02mix (approved):
LGTM

commands/upgrade/cluster/command.go
axbarsan: In `1.0.0-alpha1+clang`, `alpha1` is the pre-release information, and `clang` is the custom build metadata.
So `.Prerelease()` would return `alpha1` and `.Metadata()` would return `clang`

PR #597, opened 2020-11-16
Actually throw the error if v5 cluster upgrade fails

axbarsan:
As seen on https://gigantic.slack.com/archives/CNUDS3AMU/p1605091559107400

Currently, if the V5 cluster upgrade fails (or is blocked by an admission controller), the app doesn't actually throw an error. With this PR, it does.

### DIFF
--- a/pkg/releaseinfo/releaseinfo_test.go
+++ b/pkg/releaseinfo/releaseinfo_test.go
@@ -58,6 +58,7 @@ func TestReleaseInfo_GetReleaseData(t *testing.T) {
 			releaseVersion:            "1.0.1",
 			expectedK8sVersion:        "15.0.1",
 			expectedK8sVersionEOLDate: "2020-10-20",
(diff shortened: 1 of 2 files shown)


pkg/releaseinfo/releaseinfo_test.go
```diff
@@ -58,6 +58,7 @@ func TestReleaseInfo_GetReleaseData(t *testing.T) {
 			releaseVersion:            "1.0.1",
 			expectedK8sVersion:        "15.0.1",
 			expectedK8sVersionEOLDate: "2020-10-20",
+			expectedIsK8sVersionEOL:   true,
```
axbarsan, line 61: This is something that I missed in a previous PR, and it made a date-specific test fail

nprokopic (approved):
LGTM when CircleCI is happy :)

tuommaki (approved):

PR #598, opened 2020-11-17
Upgrade release

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/14345#issuecomment-728850246

This PR adds support for specifying the desired version to upgrade to when upgrading a tenant cluster with `gsctl upgrade`.

### DIFF
--- a/commands/upgrade/cluster/command.go
+++ b/commands/upgrade/cluster/command.go
@@ -55,6 +55,7 @@ When in doubt, please contact the Giant Swarm support team before upgrading.
 Example:
   gsctl upgrade cluster 6iec4
   gsctl upgrade cluster "Cluster name"
+  gsctl upgrade cluster "Cluster name" --release "13.0.0"
 `),
 
 		// We use PreRun for general input validation, authentication etc.
@@ -76,6 +77,7 @@ type Arguments struct {
 	AuthToken         string
 	ClusterNameOrID   string
 	Force             bool
+	Release           string
 	UserProvidedToken string
 	Verbose           bool
 }
@@ -94,6 +96,7 @@ func collectArguments(positionalArgs []string) Arguments {
 		AuthToken:         token,
 		ClusterNameOrID:   clusterID,
 		Force:             flags.Force,
+		Release:           flags.Release,
 		UserProvidedToken: flags.Token,
 		Verbose:           flags.Verbose,
 	}
@@ -113,6 +116,7 @@ func initFlags() {
 	Command.ResetFlags()
 
 	Command.Flags().BoolVarP(&flags.Force, "force", "", false, "If set, no interactive confirmation will be required (risky!).")
+	Command.Flags().StringVarP(&flags.Release, "release", "", "", "If set, the release the cluster should be upgraded to.")
 }
 
 // Prints results of our pre-validation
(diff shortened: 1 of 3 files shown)


commands/upgrade/cluster/command.go
```diff
@@ -113,6 +116,7 @@ func initFlags() {
 	Command.ResetFlags()
 
 	Command.Flags().BoolVarP(&flags.Force, "force", "", false, "If set, no interactive confirmation will be required (risky!).")
+	Command.Flags().StringVarP(&flags.Release, "release", "", "", "If set, the release the cluster should be upgraded to.")
```
axbarsan, line 119:
```suggestion
	Command.Flags().StringVarP(&flags.Release, "release", "", "", "The target release version for the upgrade. If no version is specified, the latest production version will be used.")
```

axbarsan (commented):
LGTM in general, if tested

Please also update the docs:
https://docs.giantswarm.io/reference/gsctl/upgrade-cluster/

commands/upgrade/cluster/command.go
```diff
@@ -293,6 +304,11 @@ func upgradeCluster(args Arguments) (*upgradeClusterResult, error) {
 		}
 	}
 
+	if targetRelease.Version == nil {
+		// Release was not found.
+		return nil, microerror.Maskf(errors.InvalidReleaseError, fmt.Sprintf("Release %s was not found", targetVersion))
```
fiunchinho, line 309:
```suggestion
		return nil, microerror.Maskf(errors.InvalidReleaseError, fmt.Sprintf("Can't upgrade to non existing release %s", targetVersion))
```

whites11:
One thing I'm not sure about, is that we currently don't check for downgrades.
So the use could theoretically request release 12.1.2 on a 13.0.0 cluster.
This should be handled by the admission controller (on azure) but I'm not sure if this should be blocked in the frontend directly.

WDYT @fiunchinho @axbarsan ?

fiunchinho:
> One thing I'm not sure about, is that we currently don't check for downgrades.
> So the use could theoretically request release 12.1.2 on a 13.0.0 cluster.
> This should be handled by the admission controller (on azure) but I'm not sure if this should be blocked in the frontend directly.
> 
> WDYT @fiunchinho @axbarsan ?

I think it's `admission-controller` responsibility.

axbarsan (approved):

commands/upgrade/cluster/command.go
```diff
@@ -113,6 +116,7 @@ func initFlags() {
 	Command.ResetFlags()
 
 	Command.Flags().BoolVarP(&flags.Force, "force", "", false, "If set, no interactive confirmation will be required (risky!).")
+	Command.Flags().StringVarP(&flags.Release, "release", "", "", "The target release version for the upgrade. If no version is specified, the latest production version will be used.")
```
marians, line 119: Not a blocker, just a note: "production version" is a term that could raise some question marks. In fact we haven't settled yet for a term for this. See https://github.com/giantswarm/giantswarm/issues/14171

whites11: Now that I read it again, this is false.
If no version is specified, the next "production" release will be used.
Let me rephrase it.

PR #599, opened 2020-11-30
Scalenodepoolzero

paurosello:
Towards https://github.com/giantswarm/roadmap/issues/234

In this PR I am trying to enable setting 0 as minimum for nodepools scaling commands


commands/scale/cluster/command.go
```diff
@@ -107,13 +107,13 @@ type Result struct {
 
 // getConfirmation asks the user for confirmation for scaling actions.
 func getConfirmation(args Arguments, maxBefore int, minBefore int, currentWorkers int) error {
-	if int64(currentWorkers) > args.WorkersMax && args.WorkersMax == args.WorkersMin {
+	if int64(currentWorkers) > args.WorkersMax && args.WorkersMax == *args.WorkersMin {
```
njuettner, line 110: what happens in case WorkersMin is nil? Do we need to check it here again or is the error handled before?

commands/scale/cluster/command.go
```diff
@@ -107,13 +107,13 @@ type Result struct {
...
 		confirmed := confirm.Ask(fmt.Sprintf("The cluster currently has %d worker nodes running.\nDo you want to pin the number of worker nodes to %d?", currentWorkers, args.WorkersMin))
 		if !confirmed {
 			return microerror.Mask(errors.CommandAbortedError)
 		}
 	}
-	if int64(currentWorkers) > args.WorkersMax && args.WorkersMax != args.WorkersMin {
+	if int64(currentWorkers) > args.WorkersMax && args.WorkersMax != *args.WorkersMin {
```
njuettner, line 116: same here

njuettner (commented):
just an idea, maybe we should go for consistency and also use pointer for ScalingMax. It might be easier checking all the conditions just because both ScalingMax and ScalingMin are pointers.

paurosello:
Yeah thought of that but not sure we want to be able to put max 0 in there...

Thinking on simplifying this a bit, let me git it a go

paurosello:
The latest improvements consists on simply moving to pointers when actually calling the API.

The fact that https://github.com/giantswarm/gsctl/blob/5c7820239fc68fc9552eb2751ca3c3ceda47001c/commands/update/nodepool/command.go#L67already forces us to use int64 I think now it's better.

The only new logic introduced here is `commands/update/nodepool/command.go `

marians:
@axbarsan Could you take a look here please? Thx!

axbarsan:
Regarding checking if a flag is set, here's what we do in the `scale cluster` command:

We can store whether the flag was set or not into a separate property
https://github.com/giantswarm/gsctl/blob/150fa9d904f87b7e0f1371e76990219b127c83f6/commands/scale/cluster/command.go#L92

This is how we do that actual check
https://github.com/giantswarm/gsctl/blob/150fa9d904f87b7e0f1371e76990219b127c83f6/commands/scale/cluster/command.go#L146

`cmdWorkersMaxName` being the flag name
https://github.com/giantswarm/gsctl/blob/150fa9d904f87b7e0f1371e76990219b127c83f6/commands/scale/cluster/command.go#L57

paurosello:
Thanks @axbarsan I used your method and seems to be a bit cleaner now :)

commands/update/nodepool/command.go
```diff
@@ -194,10 +196,14 @@ func updateNodePool(args Arguments) (*result, error) {
...
 			updated = true
 		}
-		if args.ScalingMin != 0 && args.ScalingMin != existingNP.Payload.Scaling.Min ||
-			args.ScalingMax != 0 && args.ScalingMax != existingNP.Payload.Scaling.Max {
+
+		//Check for non-input from user, set value to current min
+		if !args.ScalingMinSet {
+			args.ScalingMin = *existingNP.Payload.Scaling.Min
+		}
+		if (args.ScalingMin >= 0 && args.ScalingMin != *existingNP.Payload.Scaling.Min) || (args.ScalingMax >= 0 && args.ScalingMax != existingNP.Payload.Scaling.Max) {
```
axbarsan, line 204: Is it always safe to assume that `existingNP.Payload.Scaling.Min` is not `nil`? Trying to dereference it with `*existingNP.Payload.Scaling.Min` will throw an error if the pointer is nil.

commands/show/nodepool/aws.go
```diff
@@ -83,7 +83,7 @@ func formatRAMAWS(numNodes int64, details *nodespec.InstanceType) string {
 }
 
 func formatNodeScalingAWS(scaling *models.V5GetNodePoolResponseScaling) string {
-	if scaling.Min == scaling.Max {
+	if *scaling.Min == scaling.Max {
```
axbarsan, line 86: Also here, should there be a `nil` check?

commands/create/nodepool/command.go
```diff
@@ -388,10 +388,10 @@ func createNodePool(args Arguments, clusterID string, clientWrapper *client.Wrap
 		}
 	}
 
-	if args.ScalingMin != 0 || args.ScalingMax != 0 {
+	if args.ScalingMin != -1 || args.ScalingMax != 0 {
```
axbarsan, line 391: Can we also use pointers here, instead of `-1`?

commands/create/cluster/v5.go
```diff
@@ -86,8 +86,8 @@ func createAddNodePoolBody(def *types.NodePoolDefinition) *models.V5AddNodePoolR
 	}
 
 	if def.Scaling != nil {
-		if def.Scaling.Min != 0 {
-			b.Scaling.Min = def.Scaling.Min
+		if def.Scaling.Min != -1 {
```
axbarsan, line 89: Is this still correct?

commands/list/nodepools/command.go
```diff
@@ -298,7 +298,7 @@ func getOutputAWS(nps []*models.V5GetNodePoolsResponseItems) (string, error) {
 			fmt.Sprintf("%t", np.NodeSpec.Aws.UseAlikeInstanceTypes),
 			strconv.FormatInt(np.NodeSpec.Aws.InstanceDistribution.OnDemandBaseCapacity, 10),
 			strconv.FormatInt(100-np.NodeSpec.Aws.InstanceDistribution.OnDemandPercentageAboveBaseCapacity, 10),
-			strconv.FormatInt(np.Scaling.Min, 10) + "/" + strconv.FormatInt(np.Scaling.Max, 10),
+			strconv.FormatInt(*np.Scaling.Min, 10) + "/" + strconv.FormatInt(np.Scaling.Max, 10),
```
axbarsan, line 301: Can we also have `nil` checks here?

commands/scale/cluster/command.go
```diff
@@ -317,7 +317,7 @@ func scaleCluster(args Arguments) (*Result, error) {
 	scalingResult := &Result{
 		NumWorkersBefore: int(len(clusterDetails.Payload.Workers)),
 		ScalingMaxBefore: int(clusterDetails.Payload.Scaling.Max),
-		ScalingMinBefore: int(clusterDetails.Payload.Scaling.Min),
+		ScalingMinBefore: int(*clusterDetails.Payload.Scaling.Min),
```
axbarsan, line 320: Can we also have `nil` checks here, as well?

axbarsan (commented):
Final 1% 🙏🏻

paurosello:
I think I fixed the issues you mentioned

axbarsan (approved):
Amazing! Thank you for making all the changes 🤗

PR #601, opened 2020-12-03
Use new master nodes properties

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14617

This PR makes use of the new master nodes fields in the cluster creation request:
* now on Azure, if there are no supported AZs in the installation, we set the new `AvailabilityZonesUnspecified` field to `true` - works great on `germanywestcentral` 👍🏻 
* if the `--create-default-nodepool` is set to `true`, and the cluster was created with no AZs, then the nodepool is also created with no AZs

### DIFF
--- a/commands/create/cluster/v5.go
+++ b/commands/create/cluster/v5.go
@@ -11,6 +11,7 @@ import (
 	"github.com/giantswarm/gsctl/commands/errors"
 	"github.com/giantswarm/gsctl/commands/types"
 	"github.com/giantswarm/gsctl/formatting"
+	"github.com/giantswarm/gsctl/pkg/provider"
 )
 
 type definitionFromFlagsV5 struct {
@@ -18,6 +19,7 @@ type definitionFromFlagsV5 struct {
 	releaseVersion string
 	owner          string
 	isHAMaster     *bool
+	provider       string
 }
 
 // updateDefinitionFromFlagsV5 extend/overwrites a clusterDefinition based on the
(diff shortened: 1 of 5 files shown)


commands/create/cluster/v5.go
```diff
@@ -165,6 +185,13 @@ func addClusterV5(def *types.ClusterDefinitionV5, args Arguments, clientWrapper
 		}
 
 		nodePoolRequestBody := &models.V5AddNodePoolRequest{}
+
+		if response.Payload.MasterNodes != nil && len(response.Payload.MasterNodes.AvailabilityZones) < 1 {
```
axbarsan, line 189: If the created cluster has no AZs, then we also create the node pool with no AZs

commands/create/cluster/v5.go
```diff
@@ -39,9 +42,19 @@ func updateDefinitionFromFlagsV5(def *types.ClusterDefinitionV5, flags definitio
...
+			def.MasterNodes = &types.MasterNodes{
+				HighAvailability: flags.isHAMaster,
+			}
+		}
+
+	case provider.Azure:
 		def.MasterNodes = &types.MasterNodes{
-			HighAvailability: *flags.isHAMaster,
+			Azure: &types.MasterNodesAzure{
+				AvailabilityZonesUnspecified: flags.maxSupportedAZs < 1,
```
axbarsan, line 56: Default to `AvailabilityZonesUnspecified: true` when there are no supported AZs

whites11 (approved):
Looks good, thanks!

PR #602, opened 2020-12-16
Enable Azure node pool autoscaler

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14795

This PR adds node pool autoscaling support for Azure.

### DIFF
--- a/commands/create/nodepool/command.go
+++ b/commands/create/nodepool/command.go
@@ -51,7 +51,7 @@ follows:
 - Availability zones: the node pool will use 1 zone selected randomly.
 - Instance type (AWS) / VM Size (Azure): the default machine type of the installation will be
   used. Check 'gsctl info' to find out what that is.
-- Scaling settings: on AWS, the minimum will be 3 and the maximum 10 nodes, and on Azure, the minimum will be 3 and the maximum 3 nodes.
+- Scaling settings: the minimum will be 3 and the maximum 10 nodes.
 
 Examples:
 
--- a/commands/create/nodepool/command.go
+++ b/commands/create/nodepool/command.go
@@ -51,7 +51,7 @@ follows:
 - Availability zones: the node pool will use 1 zone selected randomly.
 - Instance type (AWS) / VM Size (Azure): the default machine type of the installation will be
   used. Check 'gsctl info' to find out what that is.
-- Scaling settings: on AWS, the minimum will be 3 and maximum 10 nodes, and on Azure, the node count will be 3
+- Scaling settings: on AWS, the minimum will be 3 and the maximum 10 nodes, and on Azure, the minimum will be 3 and the maximum 10 nodes.
 
 Examples:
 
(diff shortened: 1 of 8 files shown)


whites11:
Untestable on germanywestcentral. See https://github.com/giantswarm/giantswarm/issues/14915

commands/update/nodepool/command.go
```diff
@@ -139,17 +138,8 @@ func verifyPreconditions(args Arguments) error {
...
-	case provider.AWS:
-		if args.ScalingMin > args.ScalingMax && args.ScalingMax > 0 {
-			return microerror.Mask(errors.WorkersMinMaxInvalidError)
-		}
-
-	case provider.Azure:
-		if args.ScalingMin != args.ScalingMax {
-			return microerror.Maskf(errors.WorkersMinMaxInvalidError, "Provider '%s' does not support node pool autoscaling.", args.Provider)
-		}
+	if args.ScalingMin > args.ScalingMax && args.ScalingMax > 0 {
```
fiunchinho, line 141: why does `args.ScalingMax` need to be >= 0 for this to be an error?

axbarsan: I think the reason for doing that is to only throw the error if the `--nodes-max` flag is set (it has a default value of 0).

The command allows the user to set just the min scaling or just the max scaling for a node pool. (so if you want, you can only specify `--nodes-min` or `--nodes-max`). If we didn't check that the `--nodes-max` flag was set, we would throw an error if the user tried to use the `--nodes-min` flag alone.

Hope it makes sense

tuommaki (approved):
Seems good to me

commands/update/nodepool/command.go
fiunchinho: Yes, I understand now.

fiunchinho (approved):

axbarsan:
> Untestable on germanywestcentral. See [giantswarm/giantswarm#14915](https://github.com/giantswarm/giantswarm/issues/14915)

@whites11 it should work now 👍🏻

whites11:
> > Untestable on germanywestcentral. See [giantswarm/giantswarm#14915](https://github.com/giantswarm/giantswarm/issues/14915)
> 
> @whites11 it should work now 👍🏻

Tested and working perfectly, thanks a lot

commands/create/nodepool/command.go
```diff
@@ -51,7 +51,7 @@ follows:
 - Availability zones: the node pool will use 1 zone selected randomly.
 - Instance type (AWS) / VM Size (Azure): the default machine type of the installation will be
   used. Check 'gsctl info' to find out what that is.
-- Scaling settings: on AWS, the minimum will be 3 and maximum 10 nodes, and on Azure, the node count will be 3
+- Scaling settings: on AWS, the minimum will be 3 and the maximum 10 nodes, and on Azure, the minimum will be 3 and the maximum 3 nodes.
```
marians, line 54: We just discussed these defaults in SIG product and think that thexy should be the same on AWS and Azure (min=3, max=10). See https://github.com/giantswarm/roadmap/issues/251

MartaD: agreed, let's make it the same for both providers.

PR #603, opened 2020-12-17
Default NP creation to no AZs on installations that don't support them

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/14915


whites11 (approved):
LGTMIIWFY

PR #605, opened 2021-01-20
Add support for spot instances on Azure

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/15310

This PR adds spot instances support for Azure. It requires [this API branch](https://github.com/giantswarm/api/pull/1216), but it will not break without it.

The following things are added:
* A `Spot instances` row in the `show nodepool` command output. This displays if the feature is enabled or not.
  * If the feature is enabled, then the `Spot instances max price` row will display the maximum set price.
* A `Spot instances` column in the `list nodepools` command output. This will display if the feature is enabled or not.
* Two new flags added to the `create nodepool` command:
  * `--azure-spot-instances` -> enable or disable the feature
  * `--azure-spot-instances-max-price` -> set a maximum price for a VM instance

You can see a couple of previews below.

## Preview

<details>
<summary>show nodepool; spot instances off</summary>

image

</details>

<details>
<summary>show nodepool; spot instances on</summary>

image

</details>

<details>
<summary>list nodepools; spot instances off</summary>

image

</details>

<details>
<summary>list nodepools; spot instances on</summary>

image

</details>

<details>
<summary>create nodepool; spot instances off and trying to set max price</summary>

image

</details>


marians:
Just checked back how the `show nodepool` command works on AWS. There we have the `Spot instances: <number>` line, indicating how many nodes are spot instances (as a matter of status, not as a matter of desired state).

Here we have `Spot instances: <Enabled|Disabled>` as a matter of spec (desired state).

I'd like to see that discrepancy resolved somehow.

How does this actually work in Azure? Is the spot setting on Azure purely binary, so is it either on or off? When spot is on, is it guaranteed that all VMs (nodes) in fact are spot?

We could change `Spot instances: <number>` to `Spot instances count: <number>` and make it appear both in AWS and Azure. However on Azure that would only make sense the count can differ from the number of nodes.

axbarsan:
> How does this actually work in Azure? Is the spot setting on Azure purely binary, so is it either on or off? When spot is on, is it guaranteed that all VMs (nodes) in fact are spot?

Here's the Azure docs on how the feature works:
https://docs.microsoft.com/en-us/azure/virtual-machines/spot-vms

You can either have all nodes use spot instances, or none of them.

> We could change `Spot instances: <number>` to `Spot instances count: <number>` and make it appear both in AWS and Azure. However on Azure that would only make sense the count can differ from the number of nodes.

I changed AWS to use `Spot instances count`, instead.

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/giantswarm/apiextensions/v2 v2.1.0
 	github.com/giantswarm/columnize v2.0.3-0.20190718092621-cc99d98ffb29+incompatible
 	github.com/giantswarm/gscliauth v0.2.3
-	github.com/giantswarm/gsclientgen/v2 v2.0.10
+	github.com/giantswarm/gsclientgen/v2 v2.0.11-0.20210120173427-70a64e15be3c
```
axbarsan, line 14: Needs to use released version before final merge

commands/create/nodepool/command.go
```diff
@@ -113,6 +115,20 @@ Examples:
...
+  In order to use spot instances, specify the flag, like this:
+
+  gsctl create nodepool f01r4 --azure-spot-instances
+
+  Here is how you can set a maximum price that a single node pool
+  VM instance can reach before it is deallocated:
+
+  gsctl create nodepool f01r4 --azure-spot-instances --azure-spot-instances-max-price 0.00315
+
+  By setting this value to '0', the maximum price will be fixed
```
whites11, line 129:
```suggestion
  By setting this value to '0', the maximum price will be set
```

commands/create/nodepool/command.go
```diff
@@ -150,29 +166,33 @@ func initFlags() {
 	Command.Flags().BoolVarP(&flags.AWSUseAlikeInstanceTypes, "aws-use-alike-instance-types", "", false, "Use similar instance type in your node pool (AWS only). This list is maintained by Giant Swarm at the moment. Eg if you select m5.xlarge then the node pool can fall back on m4.xlarge too.")
 	Command.Flags().Int64VarP(&flags.AWSOnDemandBaseCapacity, "aws-on-demand-base-capacity", "", 0, "Number of on-demand instances that this node pool needs to have until spot instances are used (AWS only). Default is 0")
 	Command.Flags().Int64VarP(&flags.AWSSpotPercentage, "aws-spot-percentage", "", 0, "Percentage of spot instances used once the on-demand base capacity is fullfilled (AWS only). A number of 40 would mean that 60% will be on-demand and 40% will be spot instances.")
+	Command.Flags().BoolVarP(&flags.AzureSpotInstances, "azure-spot-instances", "", false, "Whether the node pool must use spot instances or on-demand.")
+	Command.Flags().Float64VarP(&flags.AzureSpotInstancesMaxPrice, "azure-spot-instances-max-price", "", 0, "Whether the node pool must use spot instances or on-demand.")
```
whites11, line 170:
```suggestion
	Command.Flags().Float64VarP(&flags.AzureSpotInstancesMaxPrice, "azure-spot-instances-max-price", "", 0, "Max bid hourly price for a single instance. 0 means on-demand price")
```

whites11 (approved):
LGTM

marians (approved):

flags/flags.go
```diff
@@ -18,6 +18,13 @@ var (
 	// to use in an node pool
 	AWSSpotPercentage int64
 
+	// AzureSpotInstances determines if spot insntances are used in a node pool.
```
axbarsan, line 21:
```suggestion
	// AzureSpotInstances determines if spot instances are used in a node pool.
```

PR #608, opened 2021-02-18
Support using kubie to set context

yasn77:
Provide option to use kubie to launch context when crating kubeconfig.

## Kind request to the PR creator: Please…

- … reference any issues this PR fixes.
- … give a few words of explanation what this PR is trying to accomplish.
- … if this PR creates changes/fixes/improvements for end users, publish a [release](https://github.com/giantswarm/gsctl/blob/master/docs/Release.md).

Thank you! ❤️

### DIFF
--- a/commands/create/kubeconfig/command.go
+++ b/commands/create/kubeconfig/command.go
@@ -129,6 +129,22 @@ func collectArguments(cmd *cobra.Command) (Arguments, error) {
 		description = "Added by user " + config.Config.Email + " using 'gsctl create kubeconfig'"
 	}
 
+	// hack..
+	// cobra sets defaults from other commands to the OutputFormat flag
+	// but we don't have "table" here, so if it's "table", set it to empty string
+	if flags.OutputFormat == "table" {
+		flags.OutputFormat = ""
+	}
+
+	if flags.UseKubie {
+		if len(cmdKubeconfigSelfContained) > 0 {
+			return Arguments{}, microerror.Maskf(errors.ConflictingFlagsError, "--kubie and --self-contained can not be used together")
+		}
+		if len(flags.OutputFormat) > 0 {
+			return Arguments{}, microerror.Maskf(errors.ConflictingFlagsError, "--kubie and --output can not be used together")
+		}
+	}
+
 	contextName := cmdKubeconfigContextName
 
 	ttl, err := util.ParseDuration(flags.TTL)
@@ -143,13 +159,6 @@ func collectArguments(cmd *cobra.Command) (Arguments, error) {
 	// apply deprecated flag if used
 	if cmd.Flags().Changed("tenant-internal") && !cmd.Flags().Changed("internal-api") {
 		flags.InternalAPI = flags.TenantInternal
-	}
-
-	// hack..
-	// cobra sets defaults from other commands to the OutputFormat flag
-	// but we don't have "table" here, so if it's "table", set it to empty string
-	if flags.OutputFormat == "table" {
-		flags.OutputFormat = ""
 	}
 
 	return Arguments{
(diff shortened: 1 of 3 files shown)


marians:
Thanks for the contribution! ❤️ 

Looks like the current implementation, when `--kubie` is used, does not set the current context in the kubectl/kubeconfig manner. Is that intended? If yes, I think it should be made more clear.

I'm not super fond of the short `-k` flag variant. So many tools in the kubernetes world start with K and I, for example, use `k` as an alias for `kubectl`. I would vote for not providing this shorthand.

yasn77:
> Looks like the current implementation, when `--kubie` is used, does not set the current context in the kubectl/kubeconfig manner. Is that intended? If yes, I think it should be made more clear.
>

Yeh, my thought was that when using Kubie, you generally don't want to touch the existing context. I also think this would be a safer approach, the context will always be clear (i.e you need to set it), when working with multiple clusters, you are less likely to run commands on the wrong cluster.

> I'm not super fond of the short `-k` flag variant. So many tools in the kubernetes world start with K and I, for example, use `k` as an alias for `kubectl`. I would vote for not providing this shorthand.

Fair point, I am ok with dropping the shorthand or how do you feel using `-b` ?

marians:
If the idea behind kubie is to not rely on the "current kubeconfig context", then I think indeed we should avoid setting the current context when `--kubie` is used.

Please take a look at the current output logic. Here is an example:

image

This needs modification for the `--kubie` case. But I don't know how exactly.

- `Switched to kubectl context 'giantswarm-<cluster>` - This isn't true. But something happened instead with kubie. How to explain that?
- The part `Whenever you want to switch to using this context: ...` might also be rather confusing, because we can assume that the kubie user wants to switch contexts some other way.

I don't think `-b` is a good shorthand for `--kubie`. Can we just stick with the long form? I'm guessing that those who prefer kubie can set an alias for themselves.

Then, there is a `--output json` option. Please check whether that needs any adaptation for the `--kubie` case.

Also please check if `--self-contained` makes sense in conjunction with `--kubie`. If not, raise a conflicting flags error.

marians (changes requested):
See my earlier comments

yasn77:
Thanks for the comments @marians 

I've refactored and updated the PR with the following:

- Error if `--kubie` is used with `--output` or `--self-contained`
- Show some output when using kubie, that informs the user of the context being used.
- Removed short option `-k`

yasn77:
@marians Not sure what to do about the code coverage. It's a small drop and I am not sure how best to add tests for kubie functionality (especially since all it does is shell out)

marians:
Thanks for taking in all the feedback! The code coverage decrease isn't a problem. It shouldn't block an approved PR.

After the output "Launching kubie with following context: ..." is printed, what exactly happens? Is kubie producing more output?

It's a tiny detail, but I wonder if that should be a success message instead like "Launched kubie with context ..." which indicates that the user is ready to go. However that would only fit in case kubie does not give such output.

marians (approved):

yasn77:
> Thanks for taking in all the feedback! The code coverage decrease isn't a problem. It shouldn't block an approved PR.
> 
> After the output "Launching kubie with following context: ..." is printed, what exactly happens? Is kubie producing more output?
> 
> It's a tiny detail, but I wonder if that should be a success message instead like "Launched kubie with context ..." which indicates that the user is ready to go. However that would only fit in case kubie does not give such output.

Kubie launches a new shell (with the kubectl context set), so it's a bit difficult to print a success message. Usually (nearly always), kubie will set the prompt, so the user should know they are now running with Kubie.

marians:
@yasn77 Can you please also provide an update for https://docs.giantswarm.io/ui-api/gsctl/create-kubeconfig/ ?

yasn77:
> @yasn77 Can you please also provide an update for https://docs.giantswarm.io/ui-api/gsctl/create-kubeconfig/ ?

Sure.. But would it be better to do it after release? BTW do I dd the release?

marians:
You can prepare the docs PR, then we do the release, then we merge the docs PR.

commands/create/kubeconfig/command.go
```diff
@@ -552,18 +591,20 @@ func createKubeconfig(ctx context.Context, args Arguments) (createKubeconfigResu
...
+			// edit kubectl config
+			if err := util.KubectlSetCluster(clusterID, result.apiEndpoint, result.caCertPath); err != nil {
+				return result, microerror.Mask(util.CouldNotSetKubectlClusterError)
+			}
+			if err := util.KubectlSetCredentials(clusterID, result.clientKeyPath, result.clientCertPath); err != nil {
+				return result, microerror.Mask(util.CouldNotSetKubectlCredentialsError)
+			}
+			if err := util.KubectlSetContext(result.contextName, clusterID); err != nil {
+				return result, microerror.Mask(util.CouldNotSetKubectlContextError)
+			}
```
marians, line 604: This means that when `--kubie` is applied, the necessary entries in the user's kubectl config file won't be created. This causes https://github.com/giantswarm/giantswarm/issues/15957

marians: I'll provide a fix

marians: https://github.com/giantswarm/gsctl/pull/610

yasn77: Thanks for fixing ❤️

PR #611, opened 2021-03-02
Enable config dir setting via env variable

kampka:
I prefer to manage project specific configuration close to the project data that relates to it, eg. `projects/foo/.aws` instead of `~/.aws`. For tools that support settings configuration via environment variables, tools like direnv make this very convenient and more importantly, an automatism, as they expose environment variables based on the shell directory that the user is in.
This PR allows `gsctl` to support this workflow better.

### DIFF
--- a/commands/info/command.go
+++ b/commands/info/command.go
@@ -261,6 +261,7 @@ func getEnvironmentVariables() map[string]string {
 	vars := []string{
 		"GSCTL_CAFILE",
 		"GSCTL_CAPATH",
+		"GSCTL_CONFIG_DIR",
 		"GSCTL_DISABLE_CMDLINE_TRACKING",
 		"GSCTL_DISABLE_COLORS",
 		"GSCTL_ENDPOINT",
--- a/commands/root.go
+++ b/commands/root.go
@@ -56,8 +56,14 @@ func init() {
 	// if it exists.
 	tokenFromEnv := os.Getenv("GSCTL_AUTH_TOKEN")
 
+	defaultConfigDir := config.DefaultConfigDirPath
+	configDirFromEnv := os.Getenv("GSCTL_CONFIG_DIR")
+	if (configDirFromEnv != "") {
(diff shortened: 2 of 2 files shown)


marians:
Thanks @kampka for the contribution!

So far I have given it only a shallow look, but plan to give it a test some time this week.

Please also add the new environment variable to this list: https://github.com/giantswarm/gsctl/blob/master/commands/info/command.go#L261-L272 - This will let the `gsctl info` command pick it up and print it if set, which simplifies debugging gsctl.

kampka:
Thanks for reviewing @marians 

> Please also add the new environment variable to this list: https://github.com/giantswarm/gsctl/blob/master/commands/info/command.go#L261-L272 - This will let the `gsctl info` command pick it up and print it if set, which simplifies debugging gsctl.

Done.

commands/info/command.go
```diff
@@ -261,6 +261,7 @@ func getEnvironmentVariables() map[string]string {
 	vars := []string{
 		"GSCTL_CAFILE",
 		"GSCTL_CAPATH",
+        "GSCTL_CONFIG_DIR",
```
ubergesundheit, line 264: Use a tab character here

```suggestion
	"GSCTL_CONFIG_DIR",
```

ubergesundheit (approved):
Apart from the tab/space issue, LGTM :+1:

commands/info/command.go
kampka: Whoops, sorry about that.

ubergesundheit (approved):
LGTM

marians (approved):

marians:
Unfortunately our CI is not configured to build for branches from forks. Does anybody know what to do?

ubergesundheit:
IMO its safe to merge, check branch https://github.com/giantswarm/gsctl/tree/config_dir_env
I pushed this branch to this repository and CI worked there.

kampka:
Thanks guys, appreciate it :+1:

PR #612, opened 2021-10-01
Added deprecated notice to commands with direct replacement

AverageMarcus:
A deprecated notice is added to those commands that have an equivalent command in the 'kubectl gs' plugin.

A general notice of deprecation of `gsctl` has been added to the readme to match the message on the docs pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 
 `gsctl` is the cross-platform command line utility to manage your Kubernetes clusters at Giant Swarm.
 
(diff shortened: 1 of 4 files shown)


marians:
This indentation looks very deep for me. Is that intentional? We should be using space instead of tabs to make it consistent for everybody.

AverageMarcus:
The existing file was using tabs rather than spaces so I followed for consistency. If we want to switch to spaces there's going to be a LOT of changes (would be best doing as its own PR)

marians (approved):
My mistake. I didn't see that you were adding a `Deprecated` field with a string value (which makes great sense!). For me it looked like you added to the `Long` string value. Never mind, LGTM!

PR #615, opened 2022-01-31
Avoid printing deprecation notices on KVM

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/765.

We currently display deprecation notices for `gsctl` commands that have a direct `kubectl gs` replacement. However, `kubectl gs` does not work for KVM installations, and so we should not show the deprecation notices there.

This PR adds a check for the installation's provider before displaying a command as deprecated. It also aligns the spelling of the word "favor" to use the American spelling.

An example of the deprecation notice shown:
Screen Shot 2022-02-01 at 10 02 47

### DIFF
--- a/commands/create/cluster/command.go
+++ b/commands/create/cluster/command.go
@@ -39,6 +39,8 @@ import (
 	"github.com/giantswarm/gsctl/commands/types"
 	"github.com/giantswarm/gsctl/flags"
 	"github.com/giantswarm/gsctl/formatting"
+	"github.com/giantswarm/gsctl/pkg/provider"
+	"github.com/giantswarm/gsctl/util"
 )
 
 // Arguments contains all possible input parameter needed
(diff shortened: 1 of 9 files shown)


marians (approved):
Nice! It fixes the problem, it seems, and you even go extra miles.

Looking at the deprecation notice as a reviewer I think that the formatting with quotes around the commands aren't ideal. We used to use color (not sure whether cyan or yellow) to highlight command literals in other contexts. That would make sense here, too. Anyway, that could be a different PR and I consider it optional.

kuosandys:
> Nice! It fixes the problem, it seems, and you even go extra miles.
> 
> Looking at the deprecation notice as a reviewer I think that the formatting with quotes around the commands aren't ideal. We used to use color (not sure whether cyan or yellow) to highlight command literals in other contexts. That would make sense here, too. Anyway, that could be a different PR and I consider it optional.

Thanks for spotting that! I checked out the UX Guidelines file and yellow is the colour reserved for commands. I've updated the deprecation notice and the screenshot above accordingly.

marians (approved):
Awesome!

